Issue #9, opened 2016-02-29
For consistency across Frost, use dasherized event names in templates


rox163:
for example, `on-change` instead of `onChange`

sandersky:
When this is addressed I'd recommend making it support both `onChange` and `on-change` so we don't have to apply a major version bump as this would be a breaking change. Then later when we need a major version bump for something else we could think about remove the camel case version `onChange`.

sglanzer:
Completely agree with @sandersky - keep the prior as a deprecated format

psbanka:
Closed by jira-gh-sync

psbanka:
Closed by jira-gh-sync

psbanka:
Closed by jira-gh-sync

PR #15, opened 2016-03-08
Consume ember-frost-component and cleanup action names

sandersky:
#MAJOR#

* Consume `ember-frost-component` for property validation instead of having own implementation
* Cleanup action names to be things like `on-change` rather than `onChange`


sandersky:
Fixes #9

addon/components/frost-bunsen-cell/component.js
```diff
@@ -1,12 +1,11 @@
 import _ from 'lodash'
-import Ember from 'ember'
 import computed, {readOnly} from 'ember-computed-decorators'
 
-import PropTypesMixin, {PropTypes} from 'ember-frost-bunsen/mixins/prop-types'
+import FrostComponent, {PropTypes} from 'ember-frost-component'
```
psbanka: Now that I look at the implementation, I am a little concerned in general with this approach. 

We are putting in a big conceptual abstraction. We now have a `FrostComponent` instead of an `Ember.Component`, and while that provides us with certain benefits, it also creates a Big Dark Area of Uncertainty for developers. "What on Earth is a `FrostComponent`?" they will ask themselves. Sure, you know and maybe I know, but the average rank-and-file developer won't know, and this will be a source of magic and vague abstraction.

In general, there is always a balance to be struck between eliminating boilerplate code and doing large and potentially confusing abstractions, and I'm worried that this may be pushing too far in the wrong direction.

sandersky: Well the alternative is to figure out how to bake this logic directly into `Ember.Component` but then I think we run the risk of people assuming `propTypes` and `getDefaultPropTypes` exist within Ember itself. In that scenario if an addon uses these concepts and for some reason the consumer is missing `ember-frost-component` or whatever we decide to call it then there components will silently fail to validate `propTypes` or apply the defaults in `getDefaultProps()`. So with that being said I think the approach I took is the cleaner way.

sandersky: I should add that when one sees `FrostComponent` being imported from `ember-frost-component` simply looking up that repository will explain everything https://github.com/ciena-frost/ember-frost-component

cstolli: Well, I'm impressed with @sandersky's work here, but I share @psbanka's concerns.  That said, I think we can use this for now, but I am wondering if there's another way to side-load this other than direct inheritance.  Is a mixin worth considering here? (not a huge fan of those, but it might be more the "ember way" here).  Also, I would really like to see this work submitted as a PR against ember core directly, regardless of what we do with it here.  Thoughts?

sandersky: Using a mixin is certainly an option but I thought `FrostComponent.extend({` looked better than `Ember.Component.extend(PropTypesMixin, {`. Guess it comes down to the more desired DX by the masses.

cstolli: @psbanka: Mixin vs. Component... does either win the cage match for you?  Any particular merits to one approach vs the other?  I'm thinking things like, "well, if it's a mixin, it's clear that's totally optional, you are still extending Ember.component, and that is better because..."

job13er: Personally, I think it's much clearer what:
```js
Ember.Component.extend(PropTypesMixin, {
```

Is giving me than

```js
FrostComponent.extend({
```

If `FrostComponent` were somehow named something like `EmberComponentWithPropTypes` it might be clear in both cases.

job13er: Even though in general I'm not a big fan of mixins, in this case it seems like the better, clearer experience.

agonza40: Typically a fan of explicitly called functions, but I think this is an OK use of mixins as well.

sglanzer: It's basically FrostComponent vs PropTypesMixin
* both require an explicit import
* FrostComponent is less clear, but can include other mixins and is upgradeable
* No-one is required to use either except for our projects

sglanzer: I'm ok with FrostComponent because it's _our_ magic and as mentioned by @sandersky it's easy to figure out what the magic is by looking at the base component (and docs I assume) and the set of mixins included

sglanzer: In either case the proptypes mixin should be available from a standalone addon

sglanzer: I guess my main argument for FrostComponent would be the ability to have multiple mixins without having to do an import for each (e.g. proptypes, ember-computed-decorators, etc.)

sandersky: I'll create `ember-prop-types` with the mixin, update `ember-frost-component` to use the mixin provided by `ember-prop-types`, and we can make this use either (it sounds like most are advocating for the mixin).

sglanzer: @sandersky Sounds good, as long as the mixin comes from an addon people have the option to completely ignore ember-frost-component, but can still choose to use it (and I'd still advocate for its usage in our components)

sandersky: [ember-prop-types](https://github.com/ciena-blueplanet/ember-prop-types) has been created

sandersky:
@psbanka does this satisfy your concerns now?

sglanzer:
Approved

psbanka:
Looks great, @sandersky – thanks for changing this up.

cstolli:
:+1:

PR #16, opened 2016-03-09
fixing addon dependency for other addons

vesper2000:
#minor#


blueprints/ember-frost-bunsen/index.js
```diff
@@ -1,6 +1,14 @@
...
-    return this.addBowerPackageToProject('z-schema', '3.16.1')
+    var blueprint = this
+    return blueprint.addBowerPackageToProject('z-schema', '3.16.1')
+      .then(function () {
+        return blueprint.addAddonsToProject({
+          packages: [
+            {name: 'ember-frost-theme', target: '^1.2.0'}
+          ]
+        })
+      })
```
sandersky: You should avoid the `blueprint` variable and instead use an arrow function like so:
```js
   afterInstall: function () {
    return this.addBowerPackageToProject('z-schema', '3.16.1')
      .then(() => {
        return blueprint.addAddonsToProject({
          packages: [
            {name: 'ember-frost-theme', target: '^1.2.0'}
          ]
        })
      })
   },
```

vesper2000: Thanks - will change.

sglanzer: Riiiight, forgot we weren't on an ancient version of node - thanks @sandersky!

PR #26, opened 2016-03-15
Remove extra computed props to avoid extra render call

sophypal:
This fixes and issue where required fields flash an error when typing initially.  This also removes an extra render call for the input components.

#PATCH#

### DIFF
--- a/addon/styles/addon.scss
+++ b/addon/styles/addon.scss
@@ -266,18 +266,3 @@ $frost-row-field-padding: 40px;
     border-top-width: 1px;
   }
 }
-
-// frost-text assumes inline-block alignment and will wrap so this override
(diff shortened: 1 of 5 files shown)


addon/styles/addon.scss
```diff
@@ -266,3 +266,18 @@ $frost-row-field-padding: 40px;
     border-top-width: 1px;
   }
 }
+
+// frost-text assumes inline-block alignment and will wrap so this override
+// uses flex to avoid the wrapping
```
sglanzer: Can we make modifications to frost-text to avoid an override?

sophypal: Gladly.

sandersky:
:+1:

PR #28, opened 2016-03-21
Add wrapper around demo to give margin-left 20px

EWhite613:
#patch#

### DIFF
--- a/tests/dummy/app/pods/application/template.hbs
+++ b/tests/dummy/app/pods/application/template.hbs
@@ -1,1 +1,3 @@
-{{outlet}}
+<div class="dummy-body">
+  {{outlet}}
+</div>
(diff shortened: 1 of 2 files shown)


tests/dummy/app/pods/application/template.hbs
```diff
@@ -1 +1,3 @@
-{{outlet}}
+<div class="dummy-body">
```
sglanzer: If you put this in the application, then it won't be present when you port into the documentation site - is that the intention?

EWhite613: @sglanzer The documentation site actually imports the application.hbs , replacing the outlet with the demo.hbs

sglanzer: Ah interesting, thanks for the clarification :)

sglanzer:
Approved

PR #29, opened 2016-03-22
improve README

sandersky:
#PATCH#

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,199 @@ ember install ember-frost-bunsen
 Coming soon…
 
 ## Examples
-Coming soon
+
+### Form View
+
+```handlebars
+{{
+  frost-bunsen-form
+  model=model
+  view=view
+}}
+```
+
+### Detail View
+
+```handlebars
+{{
+  frost-bunsen-detail
+  model=model
+  value=value
+  view=view
+}}
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,199 @@ ember install ember-frost-bunsen
...
+{
+  "type": "object",
+  "properties": {
+    "firstName": {"type": "string"},
+    "lastName": {"type": "string"}
+  }
+}
+```
+
+**View**
```
sglanzer: Is a view optional?

README.md
```diff
@@ -21,7 +21,199 @@ ember install ember-frost-bunsen
...
+  value=value
+  view=view
+}}
+```
+
+> Note: ALL values, models, and views MUST be valid [JSON](http://www.json.org/). Values are simply the data being represented in the UI which usually come directly from an API response. Models must be valid [JSON Schema](http://json-schema.org/) and views must be valid [view schema](https://github.com/ciena-frost/ember-frost-bunsen/blob/master/addon/components/validator/view-schema.js). Below we will provide examples of values, models, and views to give you a better idea of how this stuff works.
+
+### Minimal Example
+
+**Value**
```
sglanzer: Value is output, correct?  I'd suggest indicating this as a result since it looks like it's part of the interface config currently

README.md
```diff
@@ -21,7 +21,199 @@ ember install ember-frost-bunsen
...
+  "containers": [
+    {
+      "id": "main",
+      "rows": [
+        [
+          {"model": "name"},
+          {"model": "age"},
+          {"model": "married"},
+          {
+            "label": "Spouse's Name",
```
sglanzer: I'd note here that it's the first example of using view customization

sglanzer:
Approved, please consider adjustments based on the notes

README.md
sandersky: Adding the API in my next commit which outlines this.

PR #32, opened 2016-03-31
Fix password support

sandersky:
#MINOR#

### DIFF
--- a/addon/components/frost-bunsen-input-text/template.hbs
+++ b/addon/components/frost-bunsen-input-text/template.hbs
@@ -6,14 +6,25 @@
     {{/if}}
   </div>
   <div class={{inputWrapperClassName}}>
-    {{frost-text
-      class=inputClassName
-      disabled=cellConfig.disabled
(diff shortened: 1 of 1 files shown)


addon/components/frost-bunsen-input-text/template.hbs
```diff
@@ -6,14 +6,25 @@
...
-      type=inputType
-      value=renderValue
-    }}
+    {{#if (eq inputType 'password')}}
+      {{frost-password
+        class=inputClassName
+        disabled=cellConfig.disabled
+        onInput=(action "onChange")
+        placeholder=cellConfig.placeholder
+        revealable=true
```
job13er: Do we want this to be hard-coded to `true`? Or should it be configurable via the `cellConfig`?

job13er: Either way, this is better than before, so I'm gonna go ahead and approve it.

job13er:
:+1:

PR #33, opened 2016-04-01
Have mirage get namespace from environment. (For frost-guide when it …

EWhite613:
…moves to multiple sites)
#minor#


sglanzer:
Approved

PR #35, opened 2016-04-06
Redux: DO NOT MERGE

psbanka:
#MINOR#


addon/components/frost-bunsen-form/component.js
```diff
@@ -125,17 +133,28 @@ export default Ember.Component.extend(PropTypeMixin, {
...
         errors: Ember.A([]),
         warnings: Ember.A([])
       }),
-      validationResult: Ember.Object.create({
-        valid: false,
-        errors: Ember.A([]),
-        warnings: Ember.A([])
-      }),
       renderers,
       value: this.get('initialValue')
```
psbanka: any time we see value in state it should be removed.

addon/reducer.js
```diff
@@ -0,0 +1,44 @@
...
+        newState.value = unset(newState.value, bunsenId)
+      } else {
+        _.set(newState.value, bunsenId, value)
+      }
+      console.log(newState)
+      return newState
+
+    case VALIDATION_RESOLVED:
+      const validationState = {
+        value: _.cloneDeep(state.value),
```
psbanka: Don't need to cloneDeep state.value because it's not changing.

addon/components/frost-bunsen-form/component.js
```diff
@@ -125,17 +133,28 @@ export default Ember.Component.extend(PropTypeMixin, {
...
+      const onChange = this.get('onChange')
+
+      if (onChange) {
+        onChange(value)
+      }
+      const onValidation = this.get('onValidation')
+      if (onValidation) {
+        onValidation(validationResult)
+      }
+    })
```
psbanka: TODO: do `notifySubscriber()` here for all objects.

PR #36, opened 2016-04-11
Reduxification and code cleanup

sandersky:
#MAJOR#


addon/actions.js
```diff
@@ -0,0 +1,55 @@
...
+
+export function changeValue (bunsenId, value) {
+  return {
+    type: CHANGE_VALUE,
+    bunsenId,
+    value
+  }
+}
+
+export function updateValidationResults (validationResult) {
```
psbanka: (doesn't have to be done now, but: JSDOCs)

sandersky: Yeah i still need to go on a hunt and add JSDoc everywhere it is missing. Was just trying to get this 🐘 PR out to unblock @cstolli.

addon/actions.js
```diff
@@ -0,0 +1,55 @@
...
+  // Group errors by input
+  let errors = _.groupBy(validationResult.errors, 'path')
```
psbanka: I dislike the pattern of commenting lines of code like this for two reasons:
1. A line of code should be self-explanatory. If you need a comment to explain what the code does, then Something Is Wrong.
2. Comments are not testable and are often ignored when refactoring, so if a dev refactors line 19, there's a good chance they will ignore line 18 and so the comment will simply be wrong and misleading.

Here are some suggestions for how to deal with this anti-pattern:

1. Can you use different variable names so that these functions are more obvious?

```
const errorsList = validationResult.errors
const inputSelector = 'path'
let errorsByInput = _.groupBy(errorsList, inputSelector)
```

2. Write a one-line function which contains the comment, e.g.:
```
function groupErrorsByInput(errorsList) {
  return _.groupBy(errorsList, 'path')
}
...

let errors = groupErrorsByInput(validationResult.errors)
```

addon/actions.js
```diff
@@ -0,0 +1,55 @@
...
+
+  // Filter errors to just be messages not complex objects
+  errors = _.mapValues(errors, (fieldErrors, bunsenId) => _.pluck(fieldErrors, 'message'))
```
psbanka: it's okay to create new `const` objects as you go in order for this to be more self-explanatory, too, for example:
```
const errorsByInput = _.groupBy(errorsList, inputSelector)
const errorsFilteredToMessagesOnly = _.mapValues(errorsByInput, ...)
const errorsMappedToDotNotation = _.mapKeys(errorsFilteredToMessagesOnly, ...)

return {
  errors: errorsMappedToDotNotation,
  ...
}
```

addon/components/array-container.js
```diff
@@ -2,61 +2,57 @@ import 'ember-frost-bunsen/typedefs'
 
 import _ from 'lodash'
 import Ember from 'ember'
+const {A, Component} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
-
-import layout from './template'
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
-import {getLabel} from '../utils'
+import {deemberify, getLabel} from '../utils'
```
psbanka: I was hoping that with the `redux` implementation, that we could get away from `deemberify`. What are your thoughts on that?

sandersky: The reason it still exists here is because if we let the store drive the state of the array components every time you type a character in an input belonging to an array item, focus is lost as Ember recreates the array components. The only way we found around this was to have the array container maintain it's own list of array items and update that based on what the store is providing. It may still be possible to rid this code of `deemberify` but I haven't verify that.

addon/components/array-container.js
```diff
@@ -163,10 +156,39 @@ export default Ember.Component.extend(PropTypeMixin, {
...
+    }
+
+    items.forEach((item, index) => {
+      const incomingItem = value[index]
+      const stateItem = deemberify(item)
+
+      if (!_.isEqual(stateItem, incomingItem)) {
+        // const incomingItemKeys = Object.keys(incomingItem)
+        // const stateItemKeys = Object.keys(stateItem)
+        // const keysToRemove = stateItemKeys.filter((key) => incomingItemKeys.indexOf(key) < 0)
```
psbanka: Any reason to keep commented code in?

addon/components/boolean-input.js
```diff
@@ -0,0 +1,46 @@
+import _ from 'lodash'
+import computed, {readOnly} from 'ember-computed-decorators'
+import Input from './input-base'
```
psbanka: possibly "input-abstract" is a better name?

psbanka:
Okay, feel free to merge whenever you're ready. You can save any action on my comments for a subsequent PR if you like. :+1:

addon/components/array-container.js
agonza40: I think `deemberify` should only be necessary for the setting of the initial state. That way we can pass an Ember-Data record to bunsen as our seed value and still have it behave correctly.

Another improvement I think we can make to prevent re-renders is to only pass value down as an initial value OR in the case that we know we want to directly change an input's value since input components maintain their own state already.

addon/components/boolean-input.js
cstolli: InputBase would be good too, as it implies the abstract nature of the class and maps to the module name

PR #37, opened 2016-04-11
DO NOT MERGE - only for review/comments

cstolli:
@sandersky @agonza40 @psbanka

### DIFF
--- a/addon/components/frost-bunsen-input-select/component.js
+++ b/addon/components/frost-bunsen-input-select/component.js
@@ -3,13 +3,12 @@ import Ember from 'ember'
 import computed, {readOnly} from 'ember-computed-decorators'
 
 import InputMixin from 'ember-frost-bunsen/mixins/input'
+import ListInputMixin from 'ember-frost-bunsen/mixins/list-input'
 import layout from './template'
 
-import * as utils from '../utils'
-
-export default Ember.Component.extend(InputMixin, {
+export default Ember.Component.extend(InputMixin, ListInputMixin, {
   classNames: [
-    'frost-bunsen-input-text',
+    'frost-bunsen-input-select',
     'frost-field'
   ],
   layout,
@@ -23,23 +22,6 @@ export default Ember.Component.extend(InputMixin, {
   // ==========================================================================
   // Computed Properties
   // ==========================================================================
-
-  @readOnly
-  @computed('model.enum')
-  /**
-  * Get options for drop-down
-  * @param {String[]} values - drop-down values
-  * @returns {DropDownOption[]} drop-down options
-  */
-  options (values) {
-    return values.map((value) => {
-      const label = Ember.String.capitalize(
-        value.split('-').join(' ').toLowerCase()
-      )
-      return {label, value}
-    })
-  },
-
   // TODO: figure out why we can't use @readOnly (frost-text updating value property internally?)
   @computed('cellConfig.placeholder', 'state.value')
   /**
@@ -56,28 +38,6 @@ export default Ember.Component.extend(InputMixin, {
       value = placeholder
     }
     return value
-  },
-
-  /**
-   * Fetch the list of network functions from the backend and set them
-   */
-  getAsyncData () {
-    const dbStore = this.get('dbStore')
-    const query = utils.createQuery(this.get('model.query'))
-    dbStore.query(this.get('model.modelType'), query).then((resp) => {
-      const items = resp.map((resource) => {
-        const label = resource.get('label')
-        // const type = resource.get('properties.type') || 'Unknown Type'
-        const hostname = resource.get('properties.connection.hostname') || 'Unknown Hostname'
-        return {
-          label: `${label} - ${hostname}`,
-          value: resource.get('id')
-        }
-      })
-      this.set('networkFunctions', items)
-    }).catch((err) => { // eslint-disable-line handle-callback-err
-      Ember.Logger.log('Error fetching NetworkFunctions', err)
-    })
   },
 
   actions: {
--- a/addon/components/frost-bunsen-input-select/component.js
+++ b/addon/components/frost-bunsen-input-select/component.js
@@ -1,0 +1,108 @@
+import _ from 'lodash'
+import Ember from 'ember'
+import computed, {readOnly} from 'ember-computed-decorators'
+
+import InputMixin from 'ember-frost-bunsen/mixins/input'
+import layout from './template'
+
+import * as utils from '../utils'
+
+export default Ember.Component.extend(InputMixin, {
+  classNames: [
+    'frost-bunsen-input-text',
+    'frost-field'
+  ],
+  layout,
+
+  @readOnly
+  @computed('cellConfig.properties.type')
+  inputType (type) {
+    return type || 'text'
+  },
+
+  // ==========================================================================
+  // Computed Properties
+  // ==========================================================================
+
+  @readOnly
+  @computed('model.enum')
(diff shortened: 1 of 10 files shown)


addon/components/frost-bunsen-input-select/component.js
```diff
@@ -0,0 +1,68 @@
...
+export default Ember.Component.extend(InputMixin, ListInputMixin, {
+  classNames: [
+    'frost-bunsen-input-select',
+    'frost-field'
+  ],
+  layout,
+
+  @readOnly
+  @computed('cellConfig.properties.type')
+  inputType (type) {
```
sandersky: You can remove this computed property as it isn't used.

addon/components/frost-bunsen-input-select/component.js
```diff
@@ -0,0 +1,68 @@
...
+  // ==========================================================================
+  // TODO: figure out why we can't use @readOnly (frost-text updating value property internally?)
+  @computed('cellConfig.placeholder', 'state.value')
+  /**
+   * Text to render for value
+   * @param {String} placeholder - placeholder text
+   * @param {String|Boolean} value - value
+   * @returns {String} text to render
+   */
+  renderValue: function (placeholder, value) {
```
sandersky: This computed property can also be removed.

addon/components/utils.js
```diff
@@ -149,3 +149,76 @@ export function recursiveObjectCreate (object) {
...
+  const depths = startPath.split('.')
+  const valueLevels = valuePath.split('./')
+  const parentLevels = valueLevels.filter((element) => {
+    return element === '.'
+  })
+  const valueKey = valueLevels.pop()
+  const absValuePath = _.without(depths.slice(0, depths.length - parentLevels.length - 1), '', '.').join('.')
+  const absValueKey = [absValuePath, valueKey].join('.')
+  let result = _.get(obj, absValueKey)
+  return result
```
sandersky: Would just make `return _.get(obj, absValueKey)`

addon/components/utils.js
```diff
@@ -149,3 +149,76 @@ export function recursiveObjectCreate (object) {
...
+ * @returns {String} the populate queryValue
+ */
+export function parseOrchFilters (valueObj, queryValue, startPath = '') {
+  const queryFilters = queryValue.split(',')
+  const newQueryFilters = []
+  _.each(queryFilters, (param) => {
+    const pieces = param.split(':')
+    newQueryFilters.push(`${pieces[0]}:${parseOrchFilterVariables(valueObj, pieces[1], startPath)}`)
+  })
+  return newQueryFilters.join(',')
```
sandersky:
```javascript
   return queryValue.split(',').map((param) => {
     const pieces = param.split(':')
     return `${pieces[0]}:${parseOrchFilterVariables(valueObj, pieces[1], startPath)}`
   })
```

addon/mixins/list-input.js
```diff
@@ -0,0 +1,71 @@
+// import _ from 'lodash'
+import Ember from 'ember'
+// import computed, {readOnly} from 'ember-computed-decorators'
+
+import * as utils from '../components/utils'
+
+export default Ember.Mixin.create({
+  init () {
+    this._super(...arguments)
+    this.set('options', [])
```
sandersky: Since options are used by the template should this be `Ember.A([])` instead. Otherwise I assume Ember is going to magically convert it to an Ember array anyway.

sandersky:
Overall looks good.

addon/components/frost-bunsen-input-select/component.js
cstolli: Sweet, between these two getting ripped out I can completely ditch computed properties here.

addon/components/utils.js
cstolli: will do

addon/components/utils.js
cstolli: also will do

addon/mixins/list-input.js
cstolli: ok.

cstolli:
thanks @mdahl, incorporated your suggestions

addon/components/frost-bunsen-input-select/component.js
```diff
@@ -0,0 +1,68 @@
+import _ from 'lodash'
+import Ember from 'ember'
+import computed, {readOnly} from 'ember-computed-decorators'
+
+import InputMixin from 'ember-frost-bunsen/mixins/input'
+import ListInputMixin from 'ember-frost-bunsen/mixins/list-input'
+import layout from './template'
+
+export default Ember.Component.extend(InputMixin, ListInputMixin, {
```
psbanka: More than one mixin? kill me now

cstolli: There will be only the one. this will become `BaseInput.extend(ListInputMixin...`

addon/mixins/list-input.js
```diff
@@ -0,0 +1,71 @@
+// import _ from 'lodash'
+import Ember from 'ember'
+// import computed, {readOnly} from 'ember-computed-decorators'
+
+import * as utils from '../components/utils'
+
+export default Ember.Mixin.create({
+  init () {
+    this._super(...arguments)
```
psbanka: Do you get to call `this._super()` in a mixin? What does that even mean, since it doesn't participate in the inheritance chain?

cstolli: I was just thinking about that.  I'm pretty sure the mixin's `init` clobbers the mixee's `init`, and will call `_super` up the inheritance chain as normal.  This is actually an argument in favor of inheritance here.

cstolli:
closing PR.

addon/components/frost-bunsen-input-select/component.js
psbanka: I don't think this justifies the complexity. How about this alternative using composition:

```
import ListInput from 'ember-frost-bunsen/sub-components/list-input'
import layout from './template'

export default Ember.Component.extend(InputMixin, ListInputMixin, {
  classNames: [
    'frost-bunsen-input-select',
    'frost-field'
  ],
  layout,

  // ======================================================
  //                 List Input Interface
  // ======================================================
 
  init() {
    this._super(...arguments)
    this.set('options', new ListInput())
  }

  setOptions() {
    this.get('options').setOptions()
  }

  didReceiveAttrs () {
    this.super(...arguments)
    this.setOptions()
  }

  setEnumValues(values) {
    this.get('options').setEnumValues()
  }

  setAsyncDataValues (modelDef, queryDef) {
    this.get('options').setAsyncDataValues()
  }
  // Computed Properties

  @readOnly
  @computed('cellConfig.properties.type')
  inputType (type) {
    return type || 'text'
  },

 ```

psbanka: See how much easier that is to reason about? I no longer have to ask myself what crazy magic is being done in some other module, and if I have to refactor my code, I just have to look up at these five methods to know what I have to do. The burden placed on me as a developer to know what hooks are in this module has been wiped away!

sandersky: > I no longer have to ask myself what crazy magic is being done…

We are still in the world of Ember so asking yourself what crazy magic is going on may still occur. :trollface:

addon/mixins/list-input.js
sandersky: @cstolli is correct here which is probably an argument against using a mixin over inheritance.

PR #38, opened 2016-04-11
Remove need to define onChange action in custom renderers

sandersky:
#MINOR#


agonza40:
Looks good to me. Approved.

PR #48, opened 2016-04-13
BPSO-14674 - frost-select support in ember-frost-bunsen

cstolli:
#MAJOR#
now takes 'enum' and 'query' definitions in bunsen models and generates select drop-downs from them.


cstolli:
closed/reopened for lack of #VERSION# bump tag

addon/components/abstract-input.js
```diff
@@ -102,26 +102,25 @@ export default Component.extend(PropTypeMixin, {
...
   /**
    * This should be overriden by inherited inputs to convert the value to the appropriate format
-   * @param {Boolean|String} value - value to parse
+   * @param {Boolean|String} data - value to parse
    * @returns {any} parsed value
    */
-  parseValue (value) {
-    return value
+  parseValue (data) {
+    return data.value || _.get(data, 'target.value') || data
```
sandersky: This logic appears to have been moved from the `onChange` action and I believe is broken. If `data.value` is `false` for the boolean input then the `||` will evaluate.

addon/components/container.js
```diff
@@ -43,6 +43,10 @@ export default Component.extend(PropTypeMixin, {
     return result
   },
 
+  didReceiveAttrs () {
+    console.log(arguments)
```
sandersky: No need to keep this guy hanging around.

addon/components/form.js
```diff
@@ -21,13 +21,14 @@ import validateView, {validateModel} from '../validator/index'
 import {deemberify, recursiveObjectCreate} from '../utils'
 
 const builtinRenderers = {
-  PropertyChooser: 'frost-bunsen-property-chooser'
+  'PropertyChooser': 'frost-bunsen-property-chooser',
+  'select': 'frost-bunsen-input-select',
+  'multi-select': 'frost-bunsen-input-multi-select'
```
sandersky: It seem odd we went from UpperCase to lower-case.

addon/components/input-wrapper.js
```diff
@@ -60,7 +60,7 @@ export default Component.extend(PropTypeMixin, {
         return 'frost-bunsen-input-boolean'
       default:
         if (shouldRender) {
-          throw new Error('Only "string", "number", or "boolean" fields are currently supported.')
+          throw new Error('Only "string", "number", "select", or "boolean" fields are currently supported.')
```
sandersky: Shouldn't this also include `multi-select`?

addon/components/select-input.js
```diff
@@ -0,0 +1,40 @@
...
+
+  didReceiveAttrs () {
+    this._super(...arguments)
+    const modelDef = this.get('model')
+    if (modelDef) {
+      const dbStore = this.get('dbStore')
+      const promise = listUtils.getOptions(modelDef, dbStore)
+      promise.then((opts) => {
+        this.set('options', opts)
+      })
```
sandersky: I wouldn't even bother setting the variable `promise` here.
```javascript
      listUtils.getOptions(modelDef, dbStore)
        .then((opts) => {
           this.set('options', opts)
         })
```

addon/components/select-input.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+
+  dbStore: Ember.inject.service('store'),
+
+  selectedOptions: [],
+
+  didReceiveAttrs () {
+    this._super(...arguments)
+    const modelDef = this.get('model')
+    if (modelDef) {
```
sandersky: I'd change this to:
```javascript
if (!modelDef) {
  return
}
```
so the remaining logic doesn't need to be indented as much

addon/list-utils.js
```diff
@@ -0,0 +1,69 @@
...
+  const queryDef = modelDef.query
+  let result
+  if (enumDef) {
+    result = promiseValue(getEnumValues(enumDef))
+  } else if (queryDef) {
+    result = getAsyncDataValues(modelDef, queryDef, dbStore)
+  } else {
+    result = promiseValue([])
+  }
+  return result
```
sandersky:
```javascript
export function getOptions (modelDef, dbStore) {
  const enumDef = modelDef.enum
  const queryDef = modelDef.query

  if (enumDef) {
    return promiseValue(getEnumValues(enumDef))
  }

  if (queryDef) {
   return getAsyncDataValues(modelDef, queryDef, dbStore)
  }

  return promiseValue([])
}
```

addon/utils.js
```diff
@@ -150,3 +150,73 @@ export function recursiveObjectCreate (object) {
...
+ * @param {Object} obj the object to mine
+ * @param {String} valuePath the path to the value in the object
+ * @returns {String} the value
+ */
+export function findValue (obj, valuePath, startPath = '') {
+  const depths = startPath.split('.')
+  const valueLevels = valuePath.split('./')
+  const parentLevels = valueLevels.filter((element) => {
+    return element === '.'
+  })
```
sandersky:
```javascript
const parentLevels = valueLevels.filter((element) => element === '.')
```

addon/utils.js
```diff
@@ -150,3 +150,73 @@ export function recursiveObjectCreate (object) {
...
+ * @returns {String} the populated filter
+ */
+export function parseOrchFilterVariables (valueObj, orchFilter, startPath = '') {
+  let result = orchFilter
+  if (orchFilter.indexOf('${') !== -1) {
+    const valueVariable = orchFilter.split('${')[1].split('}')[0]
+    result = findValue(valueObj, valueVariable, startPath)
+  }
+  return result
+}
```
sandersky:
```javascript
export function parseOrchFilterVariables (valueObj, orchFilter, startPath = '') {
   if (orchFilter.indexOf('${') !== -1) {
     const valueVariable = orchFilter.split('${')[1].split('}')[0]
     return findValue(valueObj, valueVariable, startPath)
   }

   return orchFilter
 }
```

app/templates/components/frost-bunsen-input-select.hbs
```diff
@@ -0,0 +1,21 @@
...
+    data=options
+    selectedValue=value
+  }}
+</div>
+{{#if errorMessage}}
+  <div class={{labelWrapperClassName}}></div>
+  <div class="error">
+    {{errorMessage}}
+  </div>
+{{/if}}
```
sandersky: I believe this should actually be
```handlebars
{{#if errorMessage}}
  <div>
    <div class={{labelWrapperClassName}}></div>
      <div class="error">
        {{errorMessage}}
     </div>
  </div>
{{/if}}
```
otherwise I believe the error message will be under the label instead of the input.

addon/components/form.js
job13er: Yeah, probably change `PropertyChooser` to `property-chooser` it was always kinda weird that I made that `PascalCase` to begin with. I was pretty much referencing a React Component at the time, but since this is `ember-frost-bunsen` and not `react-frost-bunsen` and stuff will all be `dash-case` I think it makes sense to change it to `property-chooser`

addon/list-utils.js
job13er: I would probably write it @sandersky's way too, but there are advantages to the original way, particularly if you want to debug what `result` is before it's returned.

addon/list-utils.js
```diff
@@ -0,0 +1,69 @@
...
+  return values.map((value) => {
+    const label = Ember.String.capitalize(
+      value.split('-').join(' ').toLowerCase()
+    )
+    return {label, value}
+  })
+}
+
+/**
+ * setAsyncDataValues - Fetch the list of network functions from the backend and set them
```
job13er: Here's an example of why you don't want to put the name of the function in the description. Looks like you've got the wrong name, and apparently that's not checked with our `eslint` rule for `valid-jsdoc`

addon/utils.js
```diff
@@ -150,3 +150,73 @@ export function recursiveObjectCreate (object) {
...
+  return _.get(obj, absValueKey)
+}
+
+/**
+ * populates values in an Orchestrate-style 'q' or 'p' queryParam valuePath
+ * @param {Object} valueObj the value object to mine for query values
+ * @param {String} queryValue the queryParam value to parse
+ * @returns {String} the populate queryValue
+ */
+export function parseOrchFilters (valueObj, queryValue, startPath = '') {
```
job13er: I'm surprised to see this logic here. I thought that schema authors would know whether they're talking to an orch style API or not and should just fill out the queries properly. Does it make sense for bunsen to be orch API aware? How does it decide to use this or not? I guess I'll keep looking to see where it's called.

addon/list-utils.js
```diff
@@ -0,0 +1,69 @@
...
+ * @param  {Object} modelDef the full bunsen model def for this property
+ * @param  {Object} queryDef the query obj from the model def
+ * @param  {Object} dbStore the ember-data store
+ * @returns {Promise} a promise that resolves to the list of items
+ */
+export function getAsyncDataValues (modelDef, queryDef, dbStore) {
+  const query = utils.createOrchQuery(queryDef)
```
job13er: It looks like you're assuming that all queries will be in orch style. Is that a valid assumption? So no one will ever want to use this feature with a different API? It seems like making this *only* work for APIs that use a `q` and `p` query param with `:` separated queries is quite limiting. I don't think most APIs in the wild work that way. Most of the time it'll just be regular query params, right? `&foo=bar&baz=${foo.bar.baz}`? Shouldn't we just look for instances of `${}` and replace them with the proper values instead of needing to spit on `:`? Maybe I'm reading it wrong...

addon/utils.js
```diff
@@ -150,3 +150,73 @@ export function recursiveObjectCreate (object) {
...
+  _.each(keys, (key) => {
+    // get each key/value pair
+    switch (key) {
+      case 'q':
+      case 'p':
+        populatedQuery[key] = parseOrchFilters(valueObj, query[key], startPath)
+        break
+      default:
+        populatedQuery[key] = query[key]
+    }
```
job13er: Yeah, this doesn't seem right. So, if I'm understanding this code right, the only place where one can use variable substitution in the queries is inside a `q` or `p` query. What if you want to pull another query param from somewhere else, like `resourceTypeId` in an orch API, or any query param in a non orch API? I think if you look for the `${}` pattern in *all* parts of the query and substitute it, it will make it much more useful and powerful. I could imagine even wanting to pull the parameter *name* from somewhere else (a stretch, I know but doable if you just look for and replace `${}` instead of enforcing that it only exist at the right side of a `:` in an orch-style API request.

tests/integration/components/frost-bunsen-input-select-test.js
```diff
@@ -0,0 +1,125 @@
...
+        'Resource 1',
+        'Resource 2',
+        'Resource 3'
+      ]
+      _.extend(props.model, {
+        'modelType': 'resource',
+        'labelAttribute': 'label',
+        'valueAttribute': 'id',
+        'query': {
+          resourceTypeId: 'foo.bat.bitz',
```
job13er: Can we get a test case where this is pulled from somewhere else too, like:

```js
resourceTypeId: '${../../resourceType.id}'
```

Also, why are `modelType`, `labelAttribute`, `valueAttribute` and `query` wrapped in quotes, but `resourceTypeId`, `q`, and `p` are not? I think we should be consistent and either a) not wrap any keys in quotes unless necessary, or b) wrap all keys in quotes.

addon/components/abstract-input.js
cstolli: The boolean input overrides this with a specific treatment of `parseValue` that fixes the problem.  This is needed for select input to work correctly, as it returns a straight value, not an object.

addon/components/form.js
cstolli: agreed, it seemed wrong to force authors to write `SelectComponent` when they should just be able to say `select` or `multi-select`

addon/components/input-wrapper.js
cstolli: agreed, thx

addon/components/select-input.js
cstolli: agreed, I'm a fan of the early return.

addon/list-utils.js
cstolli: @job13er that is exactly how I got in that habit, the intermediary is nice.  I do like the look of @sandersky's way, it's so clean.  hrmmmm I'll switch it and see if it bites me.

addon/utils.js
cstolli: ok, I'm not in love with this shorthand, but I'll humor it.

app/templates/components/frost-bunsen-input-select.hbs
cstolli: I'm not sure that makes sense either way, honestly.  The div with class `labelWrapperClassName` is closed with no content, so the error div will appear beneath it either way.  I'm not sure how wrapping this all in an anonymous `<div>` (which is evil) will help here?

addon/list-utils.js
cstolli: thanks, good catch.

addon/utils.js
cstolli: thanks for pointing this out, completely refactored to use tail-recursion to process the stringified query object.  way cleaner, cooler, and I deleted a whole function.

tests/integration/components/frost-bunsen-input-select-test.js
cstolli: yeah, I prefer not wrapping keys unless it's proper JSON.  cleaned this up.

app/templates/components/frost-bunsen-input-select.hbs
sandersky: It has to do with flex CSS rules wanting that extra div or something.

addon/components/abstract-input.js
```diff
@@ -132,12 +132,11 @@ export default Component.extend(PropTypeMixin, {
 
     /**
      * Handle user updating value
-     * @param {Event} e - event
+     * @param {Event} data - event
      */
-    onChange (e) {
+    onChange (data) {
```
sandersky: It seems odd to name this argument `data` when the JSDoc description for it is `event`

addon/components/detail.js
```diff
@@ -28,7 +28,9 @@ function isEmberObject (object) {
 }
 
 const builtInRenderers = {
-  PropertyChooser: 'frost-bunsen-property-chooser'
+  'property-chooser': 'frost-bunsen-property-chooser',
```
sandersky: This should constitute for another major version bump.

addon/components/input-wrapper.js
```diff
@@ -60,7 +60,7 @@ export default Component.extend(PropTypeMixin, {
         return 'frost-bunsen-input-boolean'
       default:
         if (shouldRender) {
-          throw new Error('Only "string", "number", or "boolean" fields are currently supported.')
+          throw new Error('Only "string", "number", "select", "multi-select", or "boolean" fields are currently supported.')
```
sandersky: At this point I wonder if we should do:
```
const renderers = this.get('store.renderers')
const rendererNamesInQuotes = Object.keys(renderers).map((key) => `"${key}"`)
throw new Error(`Only ${rendererNamesInQuotes.join(',')} fields are currently supported.`)
```

addon/components/abstract-input.js
cstolli: ok, I'll update the description then.  not every consumer passes an event object.

addon/components/detail.js
cstolli: sounds good.

addon/components/abstract-input.js
cstolli: eh, actually I'll roll with e since it's the action handler, thanks.

addon/components/input-wrapper.js
cstolli: probably.

addon/components/number-input.js
```diff
@@ -32,7 +32,11 @@ export default AbstractInput.extend({
    * @returns {Number} parse value
    */
   parseValue (value) {
-    const number = parseFloat(value)
-    return _.isFinite(number) ? number : null
+    let result = null
+    if (value !== undefined && value !== null) {
```
sandersky: Not sure you added this additional check as `parseFloat(null)` and `parseFloat(undefined)` both return `NaN`. You should be able to just do:
```javascript
    const number = parseFloat(value)
   return _.isFinite(number) ? number : null
```

addon/components/select-input.js
```diff
@@ -0,0 +1,42 @@
...
+export default Input.extend({
+  classNames: [
+    'frost-bunsen-input-select',
+    'frost-field'
+  ],
+
+  init () {
+    this._super(...arguments)
+    this.set('options', Ember.A([]))
+  },
```
sandersky: Instead of overriding `init` here just do the following since `Input` uses the `PropTypeMixin` from `ember-prop-types`:
```javascript
getDefaultProps () {
  return {
    options: Ember.A([])
  }
},
```

cstolli: sweet!

addon/components/number-input.js
cstolli: I was getting an error without this...

PR #51, opened 2016-04-14
Code cleanup

sandersky:
#PATCH#

* Improve dependencies when populating value from consumer
* Stop doing so many Ember.Object to POJO conversions
* Add missing breaking changes to `CHANGELOG.md`
* Start removing the use of a `state` property to hold internal state in an effort to hunt down causes of extra renders
* Other minor code cleanup


job13er:
👍

PR #54, opened 2016-04-15
Update store property when redux store changes

sandersky:
#PATCH#

### DIFF
--- a/addon/components/detail.js
+++ b/addon/components/detail.js
@@ -112,16 +112,17 @@ export default Component.extend(PropTypeMixin, {
   },
 
   @readOnly
-  @computed('allRenderers', 'renderView')
+  @computed('allRenderers', 'renderValue', 'renderView')
   /**
    * Get store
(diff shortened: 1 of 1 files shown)


addon/components/detail.js
```diff
@@ -112,16 +112,17 @@ export default Component.extend(PropTypeMixin, {
...
+  @computed('allRenderers', 'renderValue', 'renderView')
   /**
    * Get store
    * @param {Object} renderers - renderer to component mapping
+   * @param {Object} formValue - current form value
    * @param {BunsenView} view - view being rendered
    * @returns {Object} store
    */
-  store (renderers, view) {
+  store (renderers, formValue, view) {
```
agonza40: Can `formValue` be changed to `renderValue` or `value`? Calling it `formValue is a little confusing because it doesn't match up with the property it is observing.

sandersky: `formValue` is used by custom renders in the wild and changing this would mean another breaking change.

PR #56, opened 2016-04-18
Defaults

agonza40:
Added handling of defaults to redux actions. 

#PATCH#

### DIFF
--- a/addon/actions.js
+++ b/addon/actions.js
@@ -68,7 +68,7 @@ function getSchema (pathStack, model, resolveRef) {
   }
 }
 
-function fillDefaults (value, path, model, resolveRef) {
+function findDefaults (value, path, model, resolveRef) {
   let schema
 
   if (model.$ref !== undefined) {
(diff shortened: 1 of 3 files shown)


addon/actions.js
```diff
@@ -26,10 +26,90 @@ export function updateValidationResults (validationResult) {
   }
 }
 
+function schemaFromRef (definitions) {
+  function invalidPath (refPath) {
```
sandersky: It seems like it would make more sense to move this function outside of its parent function.

addon/actions.js
```diff
@@ -26,10 +26,90 @@ export function updateValidationResults (validationResult) {
...
+    return model
+  } else if (model.properties) {
+    const current = pathStack.pop()
+    return getSchema(pathStack, model.properties[current], resolveRef)
+  } else if (model.items) {
+    return getSchema(pathStack, model.items, resolveRef)
+  } else {
+    return {}
+  }
+}
```
sandersky:
```javascript
function getSchema (pathStack, model, resolveRef) {
  if (model.$ref !== undefined) {
    return resolveRef(model.$ref, resolveRef)
  }

  if (pathStack.length <= 0) {
    return model
  }

  if (model.properties) {
    const current = pathStack.pop()
    return getSchema(pathStack, model.properties[current], resolveRef)
  }

  if (model.items) {
    return getSchema(pathStack, model.items, resolveRef)
  }

  return {}
}
```

PR #61, opened 2016-04-21
Bug fixes for default and falsy-values

agonza40:
Default values should now be set correctly in the case that no value has been provided.

With inputs, falsy values were causing a fall-through condition. Input values are now explicitly checked for undefined.

#PATCH#


agonza40:
@sandersky Mind reviewing this to make sure it doesn't completely destroy what you were fixing with your last PR?

@psbanka These changes should address the issues you were seeing.

sandersky:
:+1:

psbanka:
Cool; thanks for jumping on this, @agonza40 :shipit:

PR #63, opened 2016-04-21
Fix state

sandersky:
#PATCH#

### DIFF
--- a/addon/actions.js
+++ b/addon/actions.js
@@ -107,10 +107,10 @@ function findDefaults (value, path, model, resolveRef) {
 
 export function validate (bunsenId, inputValue, renderModel, validators) {
   return function (dispatch, getState) {
-    const formValue = getState()['value']
+    let formValue = getState().value
     const previousValue = _.get(formValue, bunsenId)
 
-    if (previousValue === undefined) {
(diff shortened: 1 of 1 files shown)


addon/actions.js
```diff
@@ -120,6 +120,7 @@ export function validate (bunsenId, inputValue, renderModel, validators) {
     }
 
     dispatch(changeValue(bunsenId, inputValue))
+    formValue = getState().value
```
sandersky: This allows us to validate the new value not the previous one.

addon/actions.js
```diff
@@ -107,10 +107,10 @@ function findDefaults (value, path, model, resolveRef) {
 
 export function validate (bunsenId, inputValue, renderModel, validators) {
   return function (dispatch, getState) {
-    const formValue = getState()['value']
+    let formValue = getState().value
     const previousValue = _.get(formValue, bunsenId)
 
-    if (previousValue === undefined) {
+    if (inputValue === undefined && previousValue === undefined) {
```
sandersky: It seems we were opting for a default over the newly set value when one was provided.

sandersky:
@agonza40 would love for you to review this

PR #69, opened 2016-04-22
Add integration tests around default values and user provided values

sandersky:
#PATCH#

This adds tests around user provided values to the `frost-bunsen-form` component and defaults values specified in the model to ensure the DOM being rendered contains the expect text.


agonza40:
Approved

PR #70, opened 2016-04-22
Defaults fix

agonza40:
Fixed case where arrays with defaults were causing an error when new items are added to the array.
Added a example to demo app.
#PATCH#


sandersky:
:+1:

PR #71, opened 2016-04-25
Array auto add feature

sandersky:
#MINOR#


addon/components/array-container.js
```diff
@@ -134,6 +206,55 @@ export default Component.extend(PropTypeMixin, {
       this.notifyParentOfNewItem(newItem, index)
     },
 
+    onChange (bunsenId, value) {
+      const autoAdd = this.get('cellConfig.item.autoAdd')
+      const clearingValue = [undefined, null, ''].indexOf(value) !== -1
```
agonza40: Can we use _.contains() for cases like this? (e.g. `_.contains([undefined, null, ''], value)`)

sandersky: Seeing as it is only 3 characters shorter I prefer the native approach that doesn't depend on lodash and requires one less function call.

agonza40:
Approved

PR #77, opened 2016-04-28
Adding clean and test for clean

psbanka:
When editing a value that has a null result, bunsen can get a little stuck. We need bunsen to clean out null values from initial values when getting set for the first time.

#MINOR#

### DIFF
--- a/addon/reducer.js
+++ b/addon/reducer.js
@@ -38,6 +38,28 @@ function unset (obj, path) {
   return JSON.parse(obStr)
 }
 
+/**
+ * We want to go through a state.value object and pull out any references to null
+ * @param {Object} value - our current value POJO
+ * @returns {Object} a value cleaned of any `null`s
+ */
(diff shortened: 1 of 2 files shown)


psbanka:
@agonza40 / @sandersky can you guys take a look?

addon/reducer.js
```diff
@@ -38,14 +38,36 @@ function unset (obj, path) {
...
+ * @param {Object} value - our current value POJO
+ * @returns {Object} a value cleaned of any `null`s
+ */
+function recursiveClean (value) {
+  let output = {}
+  if (_.isArray(value)) {
+    output = []
+  }
+  _.each(value, (subValue, key) => {
+    if (!_.isEmpty(subValue) || _.isNumber(subValue)) {
```
agonza40: You're handling numbers explicitly here. Will this handle boolean values?

agonza40:
Approved

PR #78, opened 2016-04-28
Improve renderers support

### LINKED ISSUE #66: Allow Ember component template names to be used as renderer
If I have a component that can be consumed via `{{foo-bar …}}` then I want to be able to specify `"renderer": "foo-bar"` without having to specify the renderer in the mapping passed into the `renderers` property of the detail/form component.

sandersky:
#MINOR#

1. You can now override built-in renderers by overriding the following renderer keys in the `renderers` property:

  * `boolean`
  * `number`
  * `string`

2. You can now specify the name of an Ember component as the value to `renderer` in your view without having to add it to the `renderers` property/mapping.

3. Fixes select inputs to only show error messages after blur.

Resolves #66

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 5.3
+
+## Non-Breaking Changes
+
+* You can now override built-in renderers by overriding the following renderer keys in the `renderers` property/mapping: `boolean`, `number`, `string`.
+* You can now specify the name of an Ember component as the value of the `renderer` property in your view without having to add it to the `renderers` property/mapping.
+
 # 5.2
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# 5.4
```
job13er: See, we have a little problem here, in that we're manually referencing version numbers that don't exist yet. What if we merged another PR before this? We really need to automate this. I think `pr-bumper` can prepend some text to `CHANGELOG.md` from the description in the PR.

sandersky: Agreed but we haven't tackled that story yet.

PR #80, opened 2016-04-29
Conditional properties

agonza40:
Initial version of conditional properties in Bunsen models and views.
Model schemas can now use the 'conditions' keyword to specify attributes that exist in the schema under certain conditions.
#MINOR#


addon/components/detail.js
```diff
@@ -83,22 +84,22 @@ export default Component.extend(PropTypeMixin, {
...
   /**
    * Get the view to render (generate one if consumer doesn't supply a view)
    * @param {BunsenModel} model - the model schema to use to generate a view (if view is undefined)
    * @param {BunsenView} view - the view to use (if given)
    * @returns {BunsenView} the view to render
    */
   renderView (model, view) {
-    view = !_.isEmpty(view) ? view : getDefaultView(model)
-    return isEmberObject(view) ? view : recursiveObjectCreate(view)
+    view = !_.isEmpty(view) ? convertView(model, view) : getDefaultView(model)
```
sandersky: Does `convertView()` work with both POJO's and Ember Objects? If not then this is a regression as the API allows for the `view` property to be a POJO or an Ember Object.

addon/components/detail.js
```diff
@@ -137,11 +138,16 @@ export default Component.extend(PropTypeMixin, {
...
       errors,
       renderValue: value
-    })
+    }
 
+    if (!_.isEqual(this.get('reduxModel'), state.model)) {
+      newProps.reduxModel = state.model
+    }
+    this.setProperties(newProps)
+    console.log(value)
```
sandersky: Line needs removed.

addon/components/detail.js
```diff
@@ -157,9 +163,10 @@ export default Component.extend(PropTypeMixin, {
   init () {
     this._super()
 
-    const reduxStore = createStoreWithMiddleware(reducer)
+    const reduxStore = createStoreWithMiddleware(reducer, initialStore({baseModel: this.get('model')}))
 
     this.set('reduxStore', reduxStore)
+    this.set('reduxModel', reduxStore.getState().model)
```
sandersky: I guess we need to capture this property along with the store itself so we can track changes.

addon/components/section.js
```diff
@@ -30,15 +30,16 @@ export default Component.extend(PropTypeMixin, {
    */
   init () {
     this._super()
-    this.set('state', Ember.Object.create({
-      expanded: this.get('expanded') || this.get('expandedOnInitialRender')
-    }))
+    this.set('state', Ember.Object.create())
+    if (this.get('expanded') === undefined) {
+      this.set('expanded', this.get('expandedOnInitialRender'))
+    }
```
sandersky: `expanded` is a property that is to be set by the consumer of this component and thus the component shouldn't be mutating it, that is why it was on the `state` property which was meant to contain internal state.

addon/convert-schema.js
```diff
@@ -0,0 +1,200 @@
...
+  if (condition.greaterThan) {
+    isConditionMet = isConditionMet || value > condition.greaterThan
+  }
+
+  if (condition.lessThan) {
+    isConditionMet = isConditionMet || value < condition.lessThan
+  }
+
+  return isConditionMet
+}
```
sandersky:
```javascript
function meetsCondition (value, condition) {
  return (
    (condition.equals && _.isEqual(condition.equals, value)) ||
    (condition.greaterThan && value > condition.greaterThan) ||
    (condition.lessThan && value < condition.lessThan)
  )
}
```

addon/convert-schema.js
```diff
@@ -0,0 +1,200 @@
...
+
+function convertConditionalProperties (model, value, getPreviousValue) {
+  if (model.type !== 'object' && model.type !== 'array' && model.properties === undefined) {
```
sandersky:
```javascript
if (!_.includes(['array', 'object'], model.type) && model.properties === undefined) {
```

addon/convert-schema.js
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    if (model.properties) {
+      const current = pathStack.pop()
+      return getSchema(pathStack, model.properties[current])
+    }
+
+    if (model.items) {
+      return getSchema(pathStack, model.items)
+    }
```
sandersky: Shouldn't this have `return view` at the bottom to ensure it always returns something?

addon/convert-schema.js
```diff
@@ -0,0 +1,200 @@
...
+  _.each(view.containers, function (container, index) {
+    _.each(container.rows, function (row, rowIndex) {
+      _.each(row, function (rowCell, rowCellIndex) {
+        const subContainer = _.get(rowCell, 'container') || _.get(rowCell, 'item.container')
+        if (subContainer) {
+          parentContainers[subContainer] = {container: container.id, rowIndex, rowCellIndex}
+        }
+      })
+    })
+  })
```
sandersky: What do you think about the following?
```javascript
const parentContainers = {}

function parseRowCell (container, rowIndex, rowCell, rowCellIndex) {
  const subContainer = _.get(rowCell, 'container') || _.get(rowCell, 'item.container')

  if (subContainer) {
    parentContainers[subContainer] = {
      container,
      rowCellIndex,
      rowIndex
    }
  }
}

function parseRow (container, row, index) {
  _.each(row, parseRowCell.bind(this, container, index))
}

function parseContainer (container, index) {
  _.each(container.rows, parseRow.bind(this, container.id))
}

_.each(view.containers, parseContainer.bind(this))
```

addon/convert-schema.js
```diff
@@ -0,0 +1,200 @@
...
+
+  function getNextContainer (containerData) {
+    if (containerData) {
+      return _.find(view.containers, function (container) {
+        return container.id === containerData.container
+      })
+    }
+  }
```
sandersky:
```javascript
  function getNextContainer (containerData) {
    if (!containerData) {
      return
    }

    return _.find(view.containers, (container) => container.id === containerData.container)
  }
```

sandersky:
Not sure if it is because it is late but I'm finding this code hard to follow without comments. Will resume looking at it in the morning to see if that helps my comprehension level.

addon/convert-schema.js
```diff
@@ -0,0 +1,200 @@
...
+  let newView = _.cloneDeep(view)
+
+  newView.containers = _.filter(_.map(view.containers, function (container, containerIndex) {
+    const newContainer = _.clone(container)
+    newContainer.rows = _.filter(_.map(container.rows, function (row, rowIndex) {
+      row = _.filter(_.map(row, function (rowCell) {
+        const path = rowCell.model.split('.').reverse().concat(getFullPath(container.id))
+        const pathClone = _.cloneDeep(path)
+        const schema = getSchema(path, model)
+        console.log(pathClone, schema)
```
sandersky: Remove line

addon/convert-schema.js
agonza40: We will end up having more than these 3 conditions and I don't want this single conditional to get too large. I'll see if I can streamline this and get it closer to a more permanent solution.

addon/convert-schema.js
agonza40: Currently, returning `undefined` indicates a missing model/schema.

sandersky: I'dd add the following for clarification
```javascript
return undefined // undefined indicates a missing model/schema
```

addon/convert-schema.js
agonza40: Not a huge fan of early returns. I also tend not to use fat arrows when not necessary because it helps me avoid relying on `this`.

addon/convert-schema.js
```diff
@@ -0,0 +1,205 @@
...
+      nextContainer = getNextContainer(containerData)
+      pathElement = nextPathElement(containerData, nextContainer)
+    }
+    return path
+  }
+  let newView = _.cloneDeep(view)
+
+  // Filter out containers with non-existent properties
+  newView.containers = _.filter(_.map(view.containers, function (container, containerIndex) {
+    const newContainer = _.clone(container)
```
lirwin: I am used to using _.compact(['foo', undefined, null]) to remove falsy values from arrays.  So _.filter(['foo', undefined, null])  with no predicate will do the same  _.filter(collection, [predicate=_.identity])  but it wasn't as apparent to me...

addon/convert-schema.js
cstolli: agree with @sandersky here, the implied return is misleading

cstolli:
overall LGTM

addon/convert-schema.js
agonza40: The only major difference I can think of is that `\_.filter` works on objects/hashes in addition to arrays, but I have to admit I wasn't familiar with `\_.compact` so I might try to use that one in the future where applicable.

addon/components/section.js
```diff
@@ -30,15 +30,16 @@ export default Component.extend(PropTypeMixin, {
    */
   init () {
     this._super()
-    this.set('state', Ember.Object.create({
-      expanded: this.get('expanded') || this.get('expandedOnInitialRender')
-    }))
+    this.set('state', Ember.Object.create())
+    if (this.get('expanded') === undefined) {
+      this.set('state.expanded', this.get('expandedOnInitialRender'))
+    }
```
sandersky: I'm not sure I understand the change here?

addon/reducer.js
```diff
@@ -1,34 +1,40 @@
...
   }
 }
 
 const INITIAL_VALUE = {
   errors: {},
   validationResult: {warnings: [], errors: []},
-  value: null
+  value: null,
+  model: {},
+  baseModel: {}
```
sandersky: What is the difference between `model` and `baseModel`? Please add comments here to clarify this.

addon/components/section.js
agonza40: Instead of setting the publicly accessible `expanded` we're now setting the private state representation of `expanded`. If the consuming context has set `expanded` directly, then we are assuming the consumer is in control of whether or not the section is expanded. If `expanded` is set, then the `didReceiveAttrs` hook will take care of setting `state.expanded`. The check for undefined is made to limit the number of `set` calls we're making, although it _should_ work fine without it.

sandersky: If you look at the diff

PR #88, opened 2016-05-05
Add new showAllErrors setting to force error visibility immediately without user interaction

sandersky:
#MINOR#


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# 5.8
+
+## Non-Breaking Changes
+
+* **Added** new property `showAllErrors` to `frost-bunsen-form` component that when true informs bunsen to show all error messages even before there is user interaction. This property defaults to `false` making it backwards compatible.
+* **Added** new property `isRequiredError` to required errors returned by the `onValidation` action/property with the value `true` to distinguish which error messages are required error messages versus which ones are not.
```
job13er: May want to clarify what a "required error message" is, it makes it sound like some error messages are optional. What we're talking about here are error messages indicating a missing required field, it might be good to explicitly say that.

addon/components/abstract-input.js
```diff
@@ -56,9 +56,9 @@ export default Component.extend(PropTypeMixin, {
   // ==========================================================================
 
   @readOnly
-  @computed('errorMessage', 'showErrorMessage')
-  renderErrorMessage (errorMessage, showErrorMessage) {
-    if (!showErrorMessage) {
+  @computed('errorMessage', 'showErrorMessage', 'store.showAllErrors')
+  renderErrorMessage (errorMessage, showErrorMessage, showAllErrors) {
+    if (showAllErrors !== true && !showErrorMessage) {
```
job13er: Is there a reason this can't just be 
```js
if (!showAllErrors && !showErrorMessage) {
```
Why do we need the explicit check against `true`?

tests/unit/components/boolean-input-test.js
```diff
@@ -44,6 +44,30 @@ describeComponent(
       ])
     })
 
+    it('shows error messages when store.showErrorMessages and showErrorMessages are true', function () {
+      component.set('showErrorMessages', false)
+      component.set('store.showAllErrors', true)
+      expect(component.get('showErrorMessages')).to.be.truthy
+    })
```
job13er: Two issues here:
 1. The text in the `it()` doesn't match what's being tested, it says both are true, but `showErrorMessages` is starting out `false`
 1. Shouldn't we be checking something like `component.get('renderErrorMessage')` vs. `component.get('showErrorMessage')`? With the test as-is, I'm fairly certain you're not exercising anything to do with `store.showAllErrors`, you're just setting `x` to be `true` then expecting `x` to be `true` still :)

job13er: Also, it appears you're working with `showErrorMessages` (note the plural) when the actual property I think you want is `showErrorMessage`

tests/unit/components/boolean-input-test.js
```diff
@@ -44,28 +44,51 @@ describeComponent(
...
+        component.set('errorMessages', errorMessages)
+      })
+
+      describe('when store.showAllErrors is true', function () {
+        beforeEach(function () {
+          component.set('store.showAllErrors', true)
+        })
+
+        it('returns error messages when showErrorMessage is true', function () {
+          expect(component.get('renderErrorMessage')).to.not.eq(errorMessages)
```
job13er: I think we want the opposite here:

```js
expect(component.get('renderErrorMessage')).to.eql(errorMessages)
```
The reason this is passing is b/c `.eq()` does a `===` (I think)

job13er: OK, I was wrong http://chaijs.com/api/bdd/#method_equal
It's `.equal()` that is `===` and `.eql()` that is a deep equal, not sure where `.eq()` comes from 😨

tests/unit/components/boolean-input-test.js
```diff
@@ -44,6 +44,47 @@ describeComponent(
       ])
     })
 
+    describe('renderErrorMessage when error message present', function () {
```
job13er: One more thing I've got to point out, or I wouldn't be able to face myself in the morning:

We've got the same code in 4 tests 😱 

could we factor that out into a helper method that takes in the `component` in a context wrapper? basically, the method looks something like this: 
```js
  function validateShowAllErrors (ctx) {
    describe('renderErrorMessage when error message present', function () {
      let component, errorMessage

      beforeEach(function () {
        component = ctx.component
        errorMessage = 'Things are borked'
        component.set('errorMessage', errorMessage)
      })

      describe('when store.showAllErrors is true', function () {
        beforeEach(function () {
          component.set('store.showAllErrors', true)
        })

        it('returns error messages when showErrorMessage is true', function () {
          component.set('showErrorMessage', true)
          expect(component.get('renderErrorMessage')).to.eql(errorMessage)
        })

        it('returns error messages when showErrorMessage is false', function () {
          component.set('showErrorMessage', false)
          expect(component.get('renderErrorMessage')).to.eql(errorMessage)
        })
      })

      describe('when store.showAllErrors is false', function () {
        beforeEach(function () {
          component.set('store.showAllErrors', false)
        })

        it('returns error messages when showErrorMessage is true', function () {
          component.set('showErrorMessage', true)
          expect(component.get('renderErrorMessage')).to.eql(errorMessage)
        })

        it('returns null when showErrorMessage is false', function () {
          component.set('showErrorMessage', false)
          expect(component.get('renderErrorMessage')).to.be.null
        })
      })
    })
  }
```

And then you call it within your individual specs like so: 
```js
import {validateShowAllErrors} from './input-test-utils'

const ctx = {}

describeComponent(
  'frost-bunsen-input-boolean',
  'FrostBunsenInputBooleanComponent',
  {},
  function () {
    let component

    beforeEach(function () {
      component = this.subject({
        bunsenId: 'enabled',
        cellConfig: Ember.Object.create({}),
        model: {},
        onChange () {},
        store: Ember.Object.create({}),
        state: Ember.Object.create({value: true})
      })
      ctx.component = component
    })

    validateShowAllErrors(ctx)
    ...
  })
```

sandersky: I like the idea, thanks for suggesting it!

job13er:
👍

PR #91, opened 2016-05-10
Treat select enum values as literals resolves #89

### LINKED ISSUE #89: Select renderer should support labels for enum values
MCP example - The enum used for select in a bunsen form is ['SSH', 'Telnet'], however the display shows 'Ssh'. It should display 'SSH' instead. 
See https://github.com/ciena-frost/ember-frost-bunsen/blob/master/addon/list-utils.js#L40

sandersky:
#PATCH#


rox163:
:+1:

PR #93, opened 2016-05-11
BPSO-18890 disable select when dependency value is not set

nskrypnik:
#PATCH#


sophypal:
Thanks @nskrypnik.  I already started on this before I knew you did all this work.  I've incorporated some of your changes.

PR #94, opened 2016-05-12
Conditional properties fix

agonza40:
Changed how inputs with no model were being handled. Previously, we were attempting to remove unused properties from view containers. However, this was creating issues when properties were used by one part of the model, then unused by another.

Inputs are now simply omitted if their model is missing.

#PATCH#


sandersky:
:+1:

PR #95, opened 2016-05-12
Add read/write transforms to abstract input

sandersky:
#MINOR#


cstolli:
👍
I got nothing on this, nice work.  Good job being functional in applying the transforms, makes perfect sense.

PR #96, opened 2016-05-13
Add support for object transforms

sandersky:
#MINOR#

### DIFF
--- a/tests/integration/components/frost-bunsen-select-test.js
+++ b/tests/integration/components/frost-bunsen-select-test.js
@@ -1,0 +1,83 @@
+import {expect} from 'chai'
+import {describeComponent, it} from 'ember-mocha'
+import {beforeEach, describe, mocha} from 'mocha'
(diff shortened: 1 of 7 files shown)


tests/integration/components/frost-bunsen-select-test.js
```diff
@@ -0,0 +1,83 @@
...
+  ],
+  rootContainers: [{
+    container: 'main',
+    label: 'Main'
+  }],
+  type: 'form',
+  version: '1.0'
+}
+
+describeComponent(...integrationTestContext('frost-bunsen-input-select'), function () {
```
job13er: Nice, I hate those multi-line `describeComponent` openings 👍

job13er:
👍

tests/integration/components/frost-bunsen-select-test.js
sandersky: We've taken it even further in a few of the test files https://github.com/ciena-frost/ember-frost-bunsen/blob/master/tests/integration/components/frost-bunsen-container-test.js#L18

PR #97, opened 2016-05-16
Adds disabled check to the select input

sophypal:
With this PR, you should be able to disable the select if it's missing required fields.  You can still force disable the field in the view which takes precedence.

#MINOR#


addon/utils.js
```diff
@@ -195,3 +195,30 @@ export function parseVariables (valueObj, queryJSON, startPath = '') {
...
+  const query = populateQuery(value, queryDef, startPath)
+  return Object.keys(query).every((key) => {
+    const prop = query[key]
+
+    // query contains multiple key-value comma-delimited pairs
+    if (prop && (prop.indexOf(':') >= 0 || prop.indexOf(',') >= 0)) {
+      return prop.split(',').every((q) => {
+        return !_.isEmpty(q.split(':')[1])
+      })
+    }
```
job13er: I'm a little concerned about this implementation. We're basically saying here that the values being pulled from somewhere else in the value object have to be used as the value of some part of the query, and we're handling the special case of the `:` and `,` delimited format for certain APIs we interface with. I think it would be more robust if we just checked that all the values references using the `${}` notation were present in the value object. 


What if one of those values is used as the property name, instead of the property value, in the query? This solution make some assumptions about where the `${}` syntax can be used that I don't think we had before (or at least I don't think we want to have). I don't think we should need to, or want to handle the `:` and `,` syntax specially at all, anywhere. If we can come up with a general solution, it should work in that special case too, and doesn't require us to have seemingly random special cases in our code :)

job13er: Maybe `populateQuery` can be modified to `throw` if a value isn't present, and we can `catch` that here?

tests/integration/components/frost-bunsen-input-select-test.js
```diff
@@ -128,22 +126,64 @@ describeComponent(...integrationTestContext('frost-bunsen-input-select'), functi
...
+
+    describe('when query dependency is not met', function () {
+      beforeEach(function () {
+        props.model = {
+          modelType: 'resource',
+          labelAttribute: 'label',
+          valueAttribute: 'id',
+          query: {
+            resourceTypeId: '${foo.bar.someOtherProp}',
+            q: 'domainId:${domainId}'
```
job13er: Can we add a use-case where the `key` in the `query` object is driven by the `${}` syntax as well? That is, if you agree to update it to support that :)

job13er:
@sophypal looks great, just the one small suggestion to make it a bit more robust.

addon/utils.js
sophypal: The `throw` would be much better instead of handling cases where `populateQuery` gave us garbage.  I should have cleaned this up more.

addon/components/select-input.js
```diff
@@ -30,6 +32,20 @@ export default AbstractInput.extend({
   // Computed Properties
   // ==========================================================================
 
+  @readOnly
+  @computed('store.formValue')
+  disabled (value) {
+    const modelDef = this.get('model')
+    const cellConfig = this.get('cellConfig')
+    const bunsenId = this.get('bunsenId')
```
sandersky: I'm curious why you didn't just do:
```js
  @computed('bunsenId', 'cellConfig', 'model', 'store.formValue')
  disabled (bunsenId, cellConfig, modelDef, value) {
```

sophypal: I'm just very cautious about what my computed properties depend on.  As far as I know, I think I only want this property recomputed (i.e. re-rendered) when the `formValue` changes.  I suppose you can say the same for `model`, `cellConfig`, and `bunsenId`.

Well, since those props aren't modified asynchronously, maybe I'm worrying about nothing.

sandersky: I think since the values of those properties effect the output of the computed property they should be included in the dependencies. In reality they shouldn't be constantly changing and probably won't effect the end result but if for some reason they are changing the disabled functionality should properly reflect that.

PR #98, opened 2016-05-17
Fix error state of select input

sandersky:
#PATCH#

# CHANGELOG

## Non-Breaking Changes

* **Fixed** issue with error state of select inputs as `error` class wasn't being passed to the underlying select component.


sandersky:
@agonza40 self-merging

PR #102, opened 2016-05-20
Add ability to disable entire form via new disabled property

sandersky:
#MINOR#

# CHANGELOG

* **Added** new property `disabled` for disabling entire form when set to `true`.


agonza40:
Approved

PR #104, opened 2016-05-22
Prevents select renderer from fetching when any change occurs

sophypal:
#PATCH#

# CHANGELOG
* **Fixed** select inputs from fetching when query hasn't changed.


sandersky:
:+1:

PR #106, opened 2016-05-24
Add a bunch of tests

sandersky:
#PATCH#

# CHANGELOG

* **Added** a bunch of unit tests around the cell container component.
* **Added** linting of Markdown and SASS to improve quality of styles and documentation.
* **Added** missing `@readOnly` decorators to properties with TODO comments.
* **Fixed** a bug when referencing an array item in a view from a container that is relative to a parent containers model.
* **Fixed** a bug when with regex pattern treating model properties with numbers in them as array items.

### DIFF
--- a/.remarkrc
+++ b/.remarkrc
@@ -1,0 +1,13 @@
+{
+  "plugins": {
+    "lint": {
+      "list-item-indent": false,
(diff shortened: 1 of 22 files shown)


sandersky:
@cstoll self-merging

PR #112, opened 2016-05-26
Nested Array Support

sandersky:
#PATCH#

# CHANGELOG

* **Added** support for nested arrays.
* **Fixed** issue with cells not always getting sub model correctly, causing nested content to not render as expected.


cstolli:
👍

PR #113, opened 2016-05-26
Performance Fixes

sandersky:
#PATCH#

# CHANGELOG

* **Fixed** issue where `onChange()` was being called even when the form value wasn't changing.
* **Fixed** issue where `onValidation()` was being called when the form error state wasn't changing.
* **Fixed** issue where form was getting in a never ending update loop causing UI to come to a crawl.


sandersky:
@agonza40 self-merging

PR #118, opened 2016-06-07
Add autofocus property to form and make demo select filterable

### LINKED ISSUE #87: deal with a cancel click and not show validation errors
open a dialog and the first input gets focus. when you dismiss the dialog, via cancel button, that triggers the validation because the Input loses focus, which then shows the error message for a fraction of a second before the dialog transitions out. It should not show the validation error if the dialog is being cancelled. Maybe bunsen should accept some kind of flag from the app, indicating that the dialog is being cancelled so don't run the validations.

### LINKED ISSUE #116: Demo Documentation for frost-bunsen-form missing 'disabled' property
the demo documentation page for frost-bunsen-form is missing `disabled` property:  http://ciena-frost.github.io/ember-frost-bunsen/#/form

### LINKED ISSUE #117: autofocusing on the first field causes issue in forms used in dialogs
A dialog launch seems to cause the onBlur to trigger which triggers validation before the user has had a chance to actually fill in any field.

Could add a property to bunsen where the user can specify whether bunsen should autofocus or leave it to the consumer.

sandersky:
#MINOR#

Resolves #117
Resolves #116
Resolves #87

# CHANGELOG

* **Added** `autofocus` property to `frost-bunsen-form` to give consumer ability to determine whether or not first input should get focus on form load.
* **Fixed** `frost-bunsen-form` documentation to include `disabled` property.
* **Updated** wedding application example in demo to make countries select searchable/filterable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ ember install ember-frost-bunsen
 
 | Attribute       | Type                       | Required | Description                              |
 | --------------- | -------------------------- | -------- | ---------------------------------------- |
+| `autofocus`     | `boolean`                  | No       | Whether or not to focus on first input   |
 | `bunsenModel`   | `Ember.Object` or `object` | Yes      | Value definition                         |
 | `bunsenView`    | `Ember.Object` or `object` | No       | View definition                          |
 | `cancelLabel`   | `string`                   | No       | Text for cancel button                   |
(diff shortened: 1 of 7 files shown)


juwara0:
👍

rox163:
:+1:

PR #123, opened 2016-06-09
Fix icon deprecations

sandersky:
#PATCH#

# CHANGELOG

* **Fixed** icons to work with latest changes in `ember-frost-core` and no longer use deprecated icon paths.
* **Updated** dependencies to latest versions.


bower.json
```diff
@@ -1,25 +1,25 @@
 {
   "name": "ember-frost-bunsen",
   "dependencies": {
-    "ember": "2.5.0",
+    "ember": "^2.6.0",
```
job13er: So, we want to force all consumers to use Ember `2.6` or later? Are there changes here that require `2.6`? Or would it be safe to just make this `^2.5.0` so it still works with apps that are pinned at `2.5.0`?

sandersky: I don't believe the dependencies in `bower.json` actually get applied to the consumer. I say this because when `perfect-scrollbar` was added to the `bower.json` of `ember-frost-core` consumers were still required to add the dependency, otherwise they'd get an error.

job13er: Right, that's the whole, blueprints part of ember

job13er: I keep thinking in terms of npm deps

PR #135, opened 2016-06-22
Add support for defining conditions in definitions

job13er:
This PR adds a new #feature# allows `conditions` to be defined within the `definitions` block of a bunsen model. It also combines two component integration tests into one and fixes the test in one which was failing. I'm guessing it snuck through by not actually being run previous to some update in `ember-cli`, `ember-cli-mocha`, or `ember-mocha`. 

PR Review task: https://agile-jira.ciena.com/browse/BPSO-23018

# CHANGELOG
* **Added** ability to define `conditions` within a `definitions` block of a bunsen model
* **Fixed** integration tests for `addon/components/select-input.js`


addon/evaluate-conditions.js
```diff
@@ -1,4 +1,11 @@
+/**
+ * @module evaluate-conditions
```
job13er: I just realized this is probably *technically* a breaking change, since I'm renaming a module that could be imported by a consumer. If desired, I can add an `addon/convert-schema.js` that imports and exports the function from `addon/evaluate-conditions.js` and throws a deprecation warning.

job13er: Done. As you can see by the current contents of `addon/convert-schema.js`

sandersky: I'd think unless our README documents consuming this file (which it doesn't), then we can consider this a private API change and thus if a consumer were consuming it, the fallout is on them.

job13er: Fair enough, but really consumers are free to import anything we export, and if we change what we export, we should probably do it in a friendly way. Particularly since this code might be handy for a custom renderer, I thought it'd be nice to go the deprecation route.

sandersky:
:+1:

addon/evaluate-conditions.js
agonza40: Do we have a way of preventing imports of modules we intend to be private? If there isn't then it's kind of unfair to us to have to support usage we don't expose in the public API.

PR #136, opened 2016-06-27
Support integer JSON schema type and add custom formats

sandersky:
#MINOR#

# CHANGELOG

* **Added** support for `integer` JSON Schema type.

* **Added** the following custom formats:

  * `bgp-as` - Border gateway protocol (BGP) autonomous system (AS)
  * `date` - date
  * `hex-string` - hexadecimal strings (ie. *2a:3b*)
  * `int8` - signed 8-bit integers
  * `int16` - signed 16-bit integers
  * `int32` - signed 32-bit integers
  * `int64` - signed 64-bit integers
  * `ipv4-address` - IPv4 address (ie. *127.0.0.1*)
  * `ipv4-interface` - IPv4 interface
  * `ipv4-prefix` - IPv4 Prefix (ie. *192.168.0.0/16*)
  * `netmask` - netmask
  * `port-number` - port
  * `time` - time
  * `uint8` - unsigned 8-bit integers
  * `uint16` - unsigned 16-bit integers
  * `uint32` - unsigned 32-bit integers
  * `url` - Uniform resource locator (URL)
  * `vlan-id` - VLAN identification number


job13er:
Just to warn you, when I first started working with `z-schema` to validate the json schema, I don't think it liked `integer`, but using `number` and `multipleOf` has worked so far.

sandersky:
@job13er thanks, I'll make sure to add some tests around integer to ensure it actually works.

addon/components/array-container.js
```diff
@@ -230,7 +230,7 @@ export default Component.extend(PropTypeMixin, {
       if (autoAdd && clearingValue) {
         const arrayPath = this.get('bunsenId')
         const itemPathBits = bunsenId.replace(`${arrayPath}.`, '').split('.')
-        const itemIndex = parseInt(itemPathBits.splice(0, 1)[0])
+        const itemIndex = parseInt(itemPathBits.splice(0, 1)[0], 10)
```
sophypal: I'm curious why it was necessary to force base 10.

sandersky: > Always specify this parameter to eliminate reader confusion and to guarantee predictable behavior. Different implementations produce different results when a radix is not specified, usually defaulting the value to 10.

Source: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt)

addon/validator/custom-formats/utils.js
```diff
@@ -0,0 +1,36 @@
+export const networkMaskMax = 32
+export const networkMaskMin = 0
+
+/**
+ * Convert decimal value to binary representation
+ * @param {Number} decimal - decimal value to convert to binary
+ * @returns {String} string containing binary representation
+ */
+export function decimalToBinary (decimal) {
+  return (decimal >>> 0).toString(2)
```
sophypal: Might be a good idea comment why the shift is necessary.  It wasn't obvious at first.

sophypal:
👍

PR #137, opened 2016-06-28
Make evaluate() a little more robust

job13er:
This #patch# just makes the `evaluate()` function a little more robust.

I just came across an instance where a model schema error caused the `evaluate()` function to be called with an empty `model` property. It proved very difficult to debug, but once I let it get through the `evaluate()` function, it was properly validated and the validation error was visible and I was able to correct the typo in my model.

# CHANGELOG
 * **Fixed** `evaluate()` to no longer choke when called with `undefined` or `null`, it just returns them as-is now.


sandersky:
:+1:

PR #140, opened 2016-07-01
Disable prototype extensions

sandersky:
#PATCH#

# CHANGELOG

* **Added** `ember-disable-prototype-extensions` to ensure code works with prototype extensions disabled.
* **Fixed** some of the annoying propType warnings.


package.json
```diff
@@ -43,17 +44,18 @@
     "ember-cli-moment-shim": "2.0.0",
     "ember-cli-notifications": "^3.2.0",
     "ember-cli-showdown": "^2.5.0",
-    "ember-cli-sri": "^2.0.0",
+    "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^1.2.0",
     "ember-computed-decorators": "0.2.2",
-    "ember-data": "^2.4.0",
+    "ember-data": "^2.5.0",
+    "ember-disable-prototype-extensions": "^1.1.0",
```
job13er: Is this just to run with no prototype extensions in the demo app?

testem.js
```diff
@@ -0,0 +1,11 @@
...
+  disable_watching: true,
+  framework: 'mocha',
+  launch_in_ci: [
+    'Firefox'
+  ],
+  launch_in_dev: [
+    'Chrome'
+  ],
+  test_page: 'tests/index.html?hidepassed&coverage'
+}
```
job13er: `.js` vs. `.json` for linting?

tests/blanket-options.js
```diff
@@ -11,7 +11,7 @@ var options = {
     autostart: true,
     lcovOptions: {
       outputFile: 'coverage/lcov.info',
-      renamer (fileName) {
+      renamer: function (fileName) {
```
job13er: Curious, why are we switching to the more verbose function definition here?

package.json
sandersky: We have it listed as a `devDependency` so that the demo app and tests for bunsen run without the prototype extensions. A consumer still has prototype extensions present unless they add `ember-disable-prototype-extensions` to their dependencies.

tests/dummy/app/app.js
```diff
@@ -1,6 +1,6 @@
 import Ember from 'ember'
 const {Application} = Ember
-import Resolver from 'ember-resolver'
+import Resolver from './resolver'
```
job13er: Why are we proxying through a module that just exports the resolver from `ember-resolver`?

testem.js
sandersky: It is linted as well as provides the ability to add comments if we wish. Also new Ember addons ship with the `..js` not a `.json` so this is more in-line with the Ember way.

tests/blanket-options.js
sandersky: I randomly saw a test fail where it complained the `:` was missing here. I have no idea why a test would say that but did this for posterities sake.

tests/dummy/app/app.js
sandersky: This is the way new Ember addons do it.

job13er:
Finished review, just a couple questions where I didn't understand the reasoning behind the change. Overall looks great.

tests/dummy/app/app.js
job13er: So the reasoning is, "this is how the cool kids do it"? Do we happen to know *why* it's done that way? Seems kinda pointless unless we're gonna change the behavior. Did you see this documented somewhere? Or was there some deprecation warning or something?

tests/blanket-options.js
job13er: > I have no idea why a test would say that but did this for ~posterities~ posterity's sake.
:trollface:

sandersky:
I should note a few of the changes you mentioned happened after I ran `ember init` on the project which runs the latest Ember blueprints against it. This was my attempt to try and keep this project from straying away from the standard scaffolding as much for less potential headaches down the road with updating Ember versions.

tests/blanket-options.js
sandersky: It's early! 😛

tests/dummy/app/app.js
job13er: So, the custom resolver was generated from the `ember init`?

job13er:
Uh-oh, I just recognized a crazy :trollface: opportunity: Apparently I can `edit` comments made by others...

job13er:
I wonder why `pullapprove` doesn't always flag the PR as needing review...

job13er:
Anyway; Approved 👍

tests/dummy/app/app.js
sandersky: Exactly.

PR #141, opened 2016-07-07
Add textarea renderer

sandersky:
#MINOR#

# CHANGELOG

* **Added** new `textarea` renderer than wraps the `frost-textarea` component from `ember-frost-core`.


agonza40:
Approved

cstolli:
@agonza40 you have to put a 👍  for Travis to see it as approved.

job13er:
Actually no, it accepts both. But I guess you realized that, since you
deleted the comment.

PR #142, opened 2016-07-07
Select default

### LINKED ISSUE #65: Don't require select renderer to be specified in view

sandersky:
#MINOR#

resolves #65 

# CHANGELOG

* Made it so the `select` renderer is automatically used when you specify the `enum` or `modelType` property for a model attribute.


cstolli:
👍  LGTM

PR #144, opened 2016-07-11
Use Ember.RSVP instead of native Promise

### LINKED ISSUE #143: Incorrect promise?
Should https://github.com/ciena-frost/ember-frost-bunsen/blob/5cd59ab7c595d3d0ce7560d56ece314909d3878a/addon/actions.js#L152 be using `Ember.RSVP.all` instead of `Promise` so the Ember run loop is aware?

sandersky:
#PATCH#

Resolves #143

# CHANGELOG

* **Fixed** code to use `Ember.RSVP` instead of native `Promise`'s to make the Ember run loop aware. This fixes failing acceptance tests in a consuming application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ Bunsen will automatically validate types of most fields, and will flag
 missing fields.  We can also pass in a list of custom validation
 functions which let us add additional conditions.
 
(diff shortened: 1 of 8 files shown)


agonza40:
Approved.

PR #145, opened 2016-07-14
Make country select in demo filterable

sandersky:
#PATCH#

# CHANGELOG

* **Updated** countries select input in wedding application demo to be filterable.


job13er:
👍

PR #146, opened 2016-07-15
Ensure initial value is set on init

sophypal:
#PATCH#

# CHANGELOG

* **Fixed** Ensures an initial value is set in `reduxStore` and resolves an issue where conditions weren't being properly evaluated in the `INIT` state.


sandersky:
:+1:

PR #147, opened 2016-07-16
Fixing dirty value changes

sophypal:
#PATCH#

My previous fix for the initial value didn't take into account `Ember` objects nor did it go through the proper channels to update the store's value which should have involved the validation loop.

# CHANGELOG

**fixes** Fixes false validation warnings on the model due to dirty `value`


sandersky:
:+1:

PR #148, opened 2016-07-19
BPSO-24495

sushant1245:
#PATCH#

# CHANGELOG

* I did performance enhancement for select-input component
* I have modified 'didrecieveAttrs' & 'hasQueryChanged' method for select-input component 
* I improved performance by 50% depending on length of bunsen form

### DIFF
--- a/addon/components/select-input.js
+++ b/addon/components/select-input.js
@@ -56,7 +56,7 @@ export default AbstractInput.extend({
       return
     }
     if (this.hasQueryChanged(oldAttrs, newAttrs, modelDef.query)) {
-           // setting required variables once above condition is true
+      // setting required variables once above condition is true
       const dbStore = this.get('dbStore')
       const value = this.get('bunsenStore.formValue')
       const bunsenId = this.get('bunsenId')
@@ -78,8 +78,8 @@ export default AbstractInput.extend({
    * @returns {Boolean} true if query has been changed
    */
   hasQueryChanged (oldAttrs, newAttrs, modelQuery) {
-     // allow models that don't have query defined to pass as well as
-     // allow the options to get initially populated
+    // allow models that don't have query defined to pass as well as
+    // allow the options to get initially populated
     if (!modelQuery || !this.get('initialized')) {
       return true
     }
--- a/addon/components/select-input.js
+++ b/addon/components/select-input.js
@@ -50,21 +50,35 @@ export default AbstractInput.extend({
 
   didReceiveAttrs ({oldAttrs, newAttrs}) {
     this._super(...arguments)
+
     const modelDef = this.get('bunsenModel')
     if (!modelDef) {
       return
     }
-    const dbStore = this.get('dbStore')
-    const value = this.get('bunsenStore.formValue')
-    const bunsenId = this.get('bunsenId')
+    //on first run "this.get('initialized')" will always be 'false' & options will be set for component
+    if(!this.get('initialized')){
 
-    if (this.hasQueryChanged(oldAttrs, newAttrs, modelDef.query) &&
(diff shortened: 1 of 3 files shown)


addon/components/select-input.js
```diff
@@ -50,21 +50,24 @@ export default AbstractInput.extend({
...
-    const value = this.get('bunsenStore.formValue')
-    const bunsenId = this.get('bunsenId')
-
-    if (this.hasQueryChanged(oldAttrs, newAttrs, modelDef.query) &&
-      utils.hasValidQueryValues(value, modelDef.query, bunsenId)) {
-      listUtils.getOptions(value, modelDef, bunsenId, dbStore).then((opts) => {
-        this.set('options', opts)
-      })
+   if (this.hasQueryChanged(oldAttrs, newAttrs, modelDef.query)) {
+           //setting required variables once above condition is true
```
job13er: Looks like you're indenting is all off, which probably has something to do with the build failing.

job13er: I'd run `npm run lint` and even `npm test` locally before submitting a PR, it'll save time and show you exactly what needs attention.

job13er:
I see changes to source files, and no changes to test files... That's generally not a good sign. If the change to source files is an additional feature, there should be new tests to cover that feature. If the change to source files is a bugfix, there should be a test that was added which fails without the bugfix. In this case, for the bugfix, there should be a test added which fails without the presence of the fix and passes with the fix.

job13er:
You are also missing a scope in your PR description, so that `pr-bumper` can know how to bump the version when the PR is merged. We should really add a `CONTRIBUTING.md` and `PULL_REQUEST_TEMPLATE.md` to this repo to introduce `pr-bumper` to potential contributors...

In the meantime, take a look [here](https://github.com/ciena-blueplanet/pr-bumper#pull-requests) for more information.

sglanzer:
Can you update the title of this PR to reference the functionality fixed/added instead of a JIRA number?  The numbers aren't meaningful in a commit log

sushant1245:
There is already a test case for select-input named as "frost-bunsen-input-select-test.js" , we have only optimize the performance of didReceiveAttrs method of select-input & no change in existing functionality of component. Hence not writing any new test case.

addon/components/select-input.js
```diff
@@ -50,20 +50,23 @@ export default AbstractInput.extend({
...
-    const value = this.get('bunsenStore.formValue')
-    const bunsenId = this.get('bunsenId')
-
-    if (this.hasQueryChanged(oldAttrs, newAttrs, modelDef.query) &&
-      utils.hasValidQueryValues(value, modelDef.query, bunsenId)) {
-      listUtils.getOptions(value, modelDef, bunsenId, dbStore).then((opts) => {
-        this.set('options', opts)
-      })
+    if (this.hasQueryChanged(oldAttrs, newAttrs, modelDef.query)) {
+           // setting required variables once above condition is true
```
sandersky: Spacing is off.

addon/components/select-input.js
```diff
@@ -75,36 +78,53 @@ export default AbstractInput.extend({
    * @returns {Boolean} true if query has been changed
    */
   hasQueryChanged (oldAttrs, newAttrs, modelQuery) {
-    // allow models that don't have query defined to pass as well as
-    // allow the options to get initially populated
+     // allow models that don't have query defined to pass as well as
+     // allow the options to get initially populated
```
sandersky: Added spaces are unnecessary.

addon/components/select-input.js
sushant1245: Are you referring to to the extra space @ line 59 
"// setting required variables once above condition is true"

psbanka:
approve

PR #155, opened 2016-08-03
Replace _.pluck() with _.map()

sandersky:
#PATCH#

# CHANGELOG

* **Replace** more `_.pluck()` with `_.map()`.


TheAxeMan301:
👍

PR #157, opened 2016-08-03
Upgrade to lodash 4

sandersky:
#PATCH#

# CHANGELOG

* **Upgraded** from `lodash` 3.x to 4.x while maintaining functionality against 3.x so consumers aren't forced to upgrade.


rox163:
:+1:

PR #158, opened 2016-08-04
correct app.import call in index.js to allow consumption in engine

rox163:
#patch#


sandersky:
:+1:

sglanzer:
We should start being conscious of the ember-cli and ember versions we require for our addons - nothing urgent, but in the interest of a LTS style release this will be important

sandersky:
@sglanzer maybe we should add older versions to our `ember-try` scenarios?

sglanzer:
That's a good point, we just need to decide what our cutoff is going to be and what time range we want to stabilize over.  Going to be a larger discussion, but it needs to be discussed soonish

PR #162, opened 2016-08-15
Call onValidation whenever value changes

job13er:
This #fix# addresses an issue w.r.t. when `onValidation` is called

# CHANGELOG
 * **Fixed** issue where `onValidation` was only called if validation result changed. It is now called whenever the value of the form changes. This fix requires a bit of explanation. If consumers want to use a `frost-bunsen-form` to update a remote resource whenever something changes (i.e. no `Submit` button), due to the async aspect of validation, one is forced to use `onValidation` to trigger a save of the last observed value from an `onChange` call. However, if the previous value was  valid, no subsequent `onValidation` was previously made when a new value was delivered with `onChange`. This effectively meant that the only way a user could actually update the remote resource would be to first enter an invalid value, then a valid one. Not exactly the best user experience, especially in the case of free text fields where no validation exists ;)

### DIFF
--- a/addon/components/detail.js
+++ b/addon/components/detail.js
@@ -191,7 +191,7 @@ export default Component.extend(PropTypeMixin, {
       onChange(value)
     }
 
-    if ('errors' in newProps && onValidation) {
+    if (('errors' in newProps || 'renderValue' in newProps) && onValidation) {
       onValidation(validationResult)
     }
   },
--- a/tests/unit/components/detail-test.js
+++ b/tests/unit/components/detail-test.js
@@ -1,7 +1,8 @@
 import {expect} from 'chai'
 import {describeComponent} from 'ember-mocha'
-import {beforeEach, it} from 'mocha'
+import {beforeEach, afterEach, describe, it} from 'mocha'
 import {PropTypes} from 'ember-prop-types'
+import sinon from 'sinon'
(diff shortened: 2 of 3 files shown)


addon/components/detail.js
```diff
@@ -191,7 +191,7 @@ export default Component.extend(PropTypeMixin, {
       onChange(value)
     }
 
-    if ('errors' in newProps && onValidation) {
+    if (('errors' in newProps || 'renderValue' in newProps) && onValidation) {
```
job13er: Here's the only actual change in functionality

tests/unit/components/detail-test.js
```diff
@@ -70,5 +76,163 @@ describeComponent(
...
+        it('should not update errors in properties', function () {
+          expect(newProps.errors).not.to.be.defined
+        })
+
+        it('should fire onChange', function () {
+          expect(changeHandler.lastCall.args).to.be.eql([newValue])
+        })
+
+        it('should fire onValidation', function () {
+          expect(validationHandler.lastCall.args).to.be.eql([{errors: []}])
```
job13er: And here's the only actual test for the new functionality. The rest were just tests to ensure existing functionality.

job13er:
@sandersky Here's that change we were talking about on Friday.

sandersky:
:+1:

PR #164, opened 2016-08-22
Add utility methods for generating facet views

sandersky:

# CHANGELOG

* **Added** the following utility methods for easily generating bunsen views for [ember-frost-object-browser](https://github.com/ciena-frost/ember-frost-object-browser) facets: `generateFacetCell`, `generateFacetView`, and `generateLabelFromModel`.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,9 @@
+**This project uses [semver](http://semver.org), please check the scope of this pr:**
+
+- [ ] #patch# - backwards-compatible bug fix
+- [ ] #minor# - adding functionality in a backwards-compatible manner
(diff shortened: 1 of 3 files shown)


agonza40:
Is the plan to remove these once we've fully deprecated version 1 views? I'm a little hesitant to keep around code in an open source repo that has to do with a specific use case.

Also, it's probably good practice to start naming our files something other than `utils.js`

sandersky:
@agonza40 this code is useful in open source as the use case is for `ember-frost-object-browser` which is open source as well. I plan to add future documentation of using `ember-frost-bunsen` for facets in `ember-frost-object-browser` which will include usage of these utilities.

agonza40:
Hmm... if we are going to provide utilities for generating views, we should probably group them into their own folder under something like `view-generation`. But for now...
👍

PR #167, opened 2016-08-24
Ensure addon works with Ember 2.2+

sandersky:

# CHANGELOG

* **Fixed** addon so it works with older versions of Ember all the way back to Ember `2.2`. Note: this means consumers now need `ember-getowner-polyfill` which they will get if they rerun the blueprints upon updating.


sandersky:
:+1:

PR #172, opened 2016-08-26
Improve logic around when to mark view cells as required

### LINKED ISSUE #171: Smarter required text
**Model**

```json
{
  "properties": {
    "foo": {
      "properties": {
        "bar": {
          "properties": {
            "spam": {
              "type": "string"
            }
          },
          "type": "object"
        },
        "baz": {
          "type": "string"
        }
      },
        "required": ["baz"],
      "type": "object"
    }
  },
  "type": "object"
}
```

**View**

```json
{
  "cells": [
    {
      "children": [
        {
          "model": "bar.spam"
        }
      ],
      "label": "Foo",
      "model": "foo"
    }
  ],
  "type": "form",
  "version": "2.0"
}
```

> Currently the `Required` text shows up under the label `Foo` because `foo.baz` is required even though we aren't rendering an input for `baz`.

Instead of blindly using the `model` attribute we should determine if any of the `children` are required when deciding whether or not a parent cell is required.

sandersky:

Resolves #171 
# CHANGELOG

* **Fixed** bug where under certain scenarios parent view cells were showing as required when they shouldn't have been.

### DIFF
--- a/addon/components/cell.js
+++ b/addon/components/cell.js
@@ -4,33 +4,12 @@ const {Component} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-cell'
+import {getMergedConfig, isRequired} from 'ember-frost-bunsen/utils'
 
 import {
-  doesModelContainRequiredField,
   getSubModel,
   getModelPath
 } from 'bunsen-core/utils'
-
-const assign = Ember.assign || Object.assign || Ember.merge
-
-/**
- * Get merged definition for current cell
- * @param {BunsenCell} cellConfig - current cell
- * @param {BunsenCell[]} cellDefinitions - list of cell definitions
- * @returns {BunsenCell} current merged cell definition
- */
-function getMergedConfig (cellConfig, cellDefinitions) {
-  if (!cellConfig.extends) {
-    return _.cloneDeep(cellConfig)
-  }
-
-  const superCell = getMergedConfig(cellDefinitions[cellConfig.extends], cellDefinitions)
-  const mergedConfig = assign(superCell, cellConfig)
-
-  delete mergedConfig.extends
-
-  return mergedConfig
-}
 
 /**
  * Return path without an index at the end
@@ -76,7 +55,7 @@ export default Component.extend(PropTypeMixin, {
   /**
    * Get definition for current cell
    * @param {BunsenCell} cellConfig - current cell
-   * @param {BunsenCell[]} cellDefinitions - list of cell definitions
+   * @param {Object<String, BunsenCell>} cellDefinitions - list of cell definitions
    * @returns {BunsenCell} current cell definition
    */
   mergedConfig (cellConfig, cellDefinitions) {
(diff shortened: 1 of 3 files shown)


addon/components/cell.js
```diff
@@ -4,34 +4,13 @@ const {Component} = Ember
...
 
-const assign = Ember.assign || Object.assign || Ember.merge
-
-/**
- * Get merged definition for current cell
- * @param {BunsenCell} cellConfig - current cell
- * @param {BunsenCell[]} cellDefinitions - list of cell definitions
- * @returns {BunsenCell} current merged cell definition
- */
-function getMergedConfig (cellConfig, cellDefinitions) {
```
sandersky: Simply moved this method to the utils file.

addon/components/cell.js
```diff
@@ -76,22 +55,24 @@ export default Component.extend(PropTypeMixin, {
   /**
    * Get definition for current cell
    * @param {BunsenCell} cellConfig - current cell
-   * @param {BunsenCell[]} cellDefinitions - list of cell definitions
+   * @param {Object<String, BunsenCell>} cellDefinitions - list of cell definitions
```
psbanka: Haven't seen this notation before. Does this mean you have a simple dictionary of the shape:
`{'foo': bunsenCellOne, 'bar': bunsenCellTwo, etc...}` ?

addon/utils.js
```diff
@@ -1,6 +1,9 @@
 import _ from 'lodash'
 import Ember from 'ember'
-const {A} = Ember
+const {A, get} = Ember
+import {getModelPath} from 'bunsen-core/utils'
+
+const assign = Ember.assign || Object.assign || Ember.merge
```
psbanka: Odd. Is this to support old versions of Ember? Also: Is there any reason to expect that `Object.assign` will not be available?

addon/components/cell.js
sandersky: That is correct. I wasn't really sure how to properly annotate that and really didn't want to just have `Object` as that doesn't tell you much.

addon/utils.js
sandersky: Yes and `Object.assign` isn't available in any version of IE according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign. I believe it also isn't available in the version of Firefox Travis uses.

addon/utils.js
```diff
@@ -97,6 +100,79 @@ export function generateFacetView (facets) {
...
+ * @returns {Boolean} whether or not last path is required
+ */
+export function isLastSegmentRequired (path, bunsenModel) {
+  const segments = path.split('.')
+  const lastSegment = segments.pop()
+
+  // Make sure we get the correct bunsen model for nested properties
+  if (segments.length !== 0) {
+    bunsenModel = get(bunsenModel, getModelPath(segments.join('.'))) || bunsenModel
+  }
```
psbanka: In general, I dislike seeing comments like line 136 and 141 in code because (1) if the code isn't readable enough to require a comment, then that's a code smell, and (2) comments like this are not maintained to the same level of attention as the code itself is maintained. In other words, you could easily see some developer coming in later and refactoring this code but leaving that comment (because most devs just ignore comments when working on code), and you could easily see that comment in six months being totally irrelevant and wrong. Instead of putting a comment like that in the code, I find it generally better to factor out a function with the same name as the comment.

For example, what about:
```
function getCorrectBunsenModelForNestedProperties(bunsenModel, segments) {
  if (segments.length !== 0) {
    bunsenModel = get(bunsenModel, getModelPath(segments.join('.'))) || bunsenModel
  }
  return bunsenModel
}
....

export function isLastSegmentRequired (path, bunsenModel) {
  const segments = path.split('.')
  const lastSegment = segments.pop()
  bunsenModel = getCorrectBunsenModelForNestedProperties(bunsenModel, segments)
  ...
}
```

addon/utils.js
psbanka: Isn't babel polyfilling it for us, though?

psbanka:
APPROVED. There are a couple small issues I raised, but nothing to prevent a merge.

addon/utils.js
sandersky: Nope. We'd have to configure it to do so in our ember config.

addon/utils.js
sandersky: Fair enough, I actually plan to try and find the reason I need this or condition in the first place and cleanup code elsewhere so I can remove it.

PR #181, opened 2016-09-08
Performance fix

job13er:

This effectively stopped bunsen from re-rendering every single input whenever the `formValue` changed. Now only those inputs which opt-in to finding out about `formValue` changes are re-rendered (`select` and `hidden`) inputs are the only built-ins with this behavior. 

# CHANGELOG

## Breaking changes
 * **Removed** - `bunsenStore` from properties passed to a custom renderer. Some of the properties that used to be provided in `bunsenStore` are not provided individually (see below)
 * **Added** - `bunsenView` to properties passed to a custom renderer
 * **Added** - `formDisabled` to properties passed to a custom renderer
 * **Added** - `showAllErrors` to properties passed to a custom renderer
 * **Added** - `registerForFormValueChanges` to properties passed to a custom renderer, if a custom renderer wants to know about changes to `formValue` it must now call the passed in `registerForFormValueChanges` inside `init()` and pass itself into said function `this.registerForFormValueChanges(this)`, the custom renderer must also provide a `formValueChanged` function on the component which will be called by the parent `frost-bunsen-form` or `frost-bunsen-detail` component whenever the `formValue` changes. The only parameter to the `formValueChanged` function is the new `formValue`. 

## Non-breaking changes
 * **Added** ability for the demo app to store currently selected view/model/value in the URL
 * **Added** the `hidden` renderer which allows setting a value in the form without displaying anything to the user. The value can come from the `default` in the bunsen model, or be copied from the value of some other portion of the `formValue` via the `valueRef` property. For instance ,if I want the `label` attribute to be set to what the user entered in `name`, I could specify a `valueRef` of `name`. The `valueRef` is the dotted path from the root of `formValue` to where the value should come from (not relative to the value being set).


tests/unit/components/form-test.js
```diff
@@ -1,253 +1,149 @@
...
+  })
 
-    it('actions.onTabChange() updates selectedTabIndex', function () {
-      [0, 1, 2].forEach((index) => {
-        component.actions.onTabChange.call(component, index)
-        expect(component.get('selectedTabIndex')).to.eql(index)
-      })
+  it('actions.onTabChange() updates selectedTabIndex', function () {
+    [0, 1, 2].forEach((index) => {
+      component.send('onTabChange', index)
```
job13er: This is one piece I wanted to make sure I called out. If you want to call an action on a component, you don't need to use
 ```js
component.actions.actionName.call(component, param)
``` 
you can use 
```js
component.send('actionName', param)
```

sandersky:
:+1:

job13er:
@sandersky had already approved this, so I'm gonna go ahead and Approve it too.

job13er:
👍

PR #183, opened 2016-09-09
Fix select renderer w.r.t. text filters in queries

job13er:

# CHANGELOG

## Breaking changes
 * **Stopped** automatically creating `p=label:foo` query param in select-renderer when user types `foo` This was useful for a particular API we were interfacing with early on, but never should have been in the open-source version of this component.

## Non-breaking changes
 * **Added** support for defining `$filter` in your definition of `query` within the bunsen model for a `select` input, which lets you specify what the query looks like to do text matching in query-driven select renderers. 
 * **Added** a new `eslint` rule in `tests/.eslintrc` to stop us using `expect(foo).to.be.true` or other dangling property matchers from `chai` b/c they are dangerous. For more information about why that is, ask @job13er.


sandersky:
:+1:

addon/list-utils.js
```diff
@@ -79,7 +81,9 @@ export function getAsyncDataValues (value, modelDef, bunsenId, dbStore, filter)
       }
     })
     return items
+  }, function (err) {
+    console.log(`it errored ${err}`)
```
job13er: Oops, didn't mean to leave this in here. I'm gonna get rid of it real quick.

job13er:
Approved

PR #190, opened 2016-09-21
Maintain Ember Data item's ID property when seriailizing

sandersky:

# CHANGELOG

* **Fixed** serialization of Ember Data objects to keep `id` in object so it can be referenced in a bunsen model.


sglanzer:
This is good, but I think it needs to be an option, I've seen enough conversations where people are trying to remove the auto ids

PR #193, opened 2016-09-26
Fix issue with defaults in object array item

### LINKED ISSUE #174: When an array has defaults, the first item added doesn't get the joke
As pointed out by `mstaffor` in Slack:

> quick question. for the example of the Model `Array with Defaults`, is there a reason that the first superhero added does not get supplied with the defaults, but the additional ones do ?

pasted_image_at_2016_08_31_09_22

That is certainly a bug.

sandersky:

Resolves #174 

# CHANGELOG

* **Fixed** issue where first item added to an object array wasn't getting defaults applied.

### DIFF
--- a/addon/components/array-container.js
+++ b/addon/components/array-container.js
@@ -285,22 +285,6 @@ export default Component.extend(PropTypeMixin, {
     this.set('formValue', newValue)
   },
 
-  /**
-   * Fire an initial BunsenChangeEvent to let parent know that a new item was added to the array
-   * @param {Object} item - the item being added
-   * @param {Number} index - the index of the item
-   */
-  notifyParentOfNewItem (item, index) {
-    const bunsenId = this.get('bunsenId')
-    const onChange = this.get('onChange')
-
-    if (this.get(`formValue.${bunsenId}`)) {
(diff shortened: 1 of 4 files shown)


addon/components/array-container.js
```diff
@@ -310,11 +294,12 @@ export default Component.extend(PropTypeMixin, {
     onAddItem () {
       const bunsenId = this.get('bunsenId')
       const newItem = this._getEmptyItem()
+      const onChange = this.get('onChange')
       const value = this.get('value')
       const items = _.get(value, bunsenId) || []
       const index = items.length
 
-      this.notifyParentOfNewItem(newItem, index)
+      onChange(`${bunsenId}.${index}`, newItem)
```
cbeekciena: So by calling onChange() you force a re-rendering, correct?

agonza40:
👍

addon/components/array-container.js
cbeekciena: Matt's response, as discussed offline:

Yes b/c onChange propagates back up to the detail component, which in turn hands the change event off to the redux store (in bunsen-core), which ultimately updates the form value in the detail component

PR #198, opened 2016-10-04
Prevent excessive re-rendering intermediate components

sophypal:

# CHANGELOG

* Added change-set processing to the view cells so they only propagate the form value when it has child cells that need updating.


addon/components/cell.js
```diff
@@ -23,6 +24,21 @@ export function removeIndex (path) {
...
+
+    if (current.done) {
+      break
+    }
+
+    if (iteratee(current.value) === false) {
+      break
+    }
+  }
+}
```
sandersky: Can't you rewrite this like so:
```js
export function iterateMap (iterator, iteratee) {
  while (let current = iterator.next()) {
    if (current.done || iteratee(current.value) === false) {
      break
    }
  }
}
```

addon/components/cell.js
```diff
@@ -23,6 +24,21 @@ export function removeIndex (path) {
   return /^\d+$/.test(last) ? parts.join('.') : path || ''
 }
 
+export function iterateMap (iterator, iteratee) {
```
sandersky: JSDoc

addon/components/cell.js
```diff
@@ -52,26 +68,45 @@ export default Component.extend(PropTypeMixin, {
...
+      readOnly: false,
+      propagatedValue: {},
+      propagatedValueChangeSet: null
+    }
+  },
+
+  didReceiveAttrs ({oldAttrs, newAttrs}) {
+    const valueChangeSet = this.get('valueChangeSet')
+    const oldCellConfig = _.get(oldAttrs, 'cellConfig.value')
+    const cellConfig = this.get('cellConfig')
```
sandersky: I personally like to use the prefix `new` when in `didReceiveAttrs`, in this case I'd use `newCellConfig`.

addon/components/cell.js
```diff
@@ -52,26 +68,45 @@ export default Component.extend(PropTypeMixin, {
...
+        if (isCommonAncestor(cellConfig.__dependency__, bunsenId)) {
+          isDirty = true
+          return false
+        }
+      })
+    }
+
+    if (isDirty || !_.isEqual(oldCellConfig, cellConfig)) {
+      this.set('propagatedValue', this.get('value'))
+      this.set('propagatedValueChangeSet', this.get('valueChangeSet'))
```
sandersky: When setting more than one property use `setProperties`.

```js
this.setProperties({
   propagatedValue: this.get('value'),
  propagatedValueChangeSet: this.get('valueChangeSet')
})
```

addon/components/detail.js
```diff
@@ -138,11 +152,27 @@ export default Component.extend(PropTypeMixin, {
...
+   * @param {Object[]} cells - the view cells
+   * @param {Object} cellDefinitions - the view cell definitions
+   * @returns {Object} cellConfigs with  precomputed model dependencies
+   */
   @readOnly
-  @computed('renderView')
-  cellTabs () {
-    const cells = this.get('renderView.cells')
+  @computed('renderView.cells', 'renderView.cellDefinitions')
+  precomputedCellConfig (cells, cellDefinitions) {
```
sandersky: So we are now pre-dereferencing the cells instead of the using `mergedConfig` logic I had in the cell component?

sophypal: You are correct. There was nothing non-performant with the `mergedConfig` logic you had before, I just needed to precompute it so that I know all the dependencies model before it had a chance to render things.

sandersky: I've wanted to move the logic into one place similar to what you did here anyways. :+1:

sandersky:
:+1:

sandersky:
:+1:

PR #202, opened 2016-10-07
Added enum driven checkbox-array renderer

PoplarToppler:

# CHANGELOG

* **Added** an enum driven checkbox-array renderer to provide a flat checkbox facet selection group with the ability to select multiple options.

### DIFF
--- a/addon/components/inputs/checkbox-array.js
+++ b/addon/components/inputs/checkbox-array.js
@@ -1,6 +1,9 @@
+import _ from 'lodash'
 import AbstractInput from './abstract-input'
 import computed, {readOnly} from 'ember-computed-decorators'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-checkbox-array'
+import Ember from 'ember'
+const {get} = Ember
 
 export default AbstractInput.extend({
   // == Component Properties ===================================================
@@ -17,20 +20,25 @@ export default AbstractInput.extend({
   selected: [],
 
   // == Computed Properties ====================================================
+  @readOnly
+  @computed('cellConfig')
+  size (cellConfig) {
+    return _.get(cellConfig, 'renderer.size') || 'medium'
+  },
 
   @readOnly
   @computed('bunsenModel')
   options (bunsenModel) {
-    return bunsenModel.items.enum.map((key) => {
-      return {
-        value: key
-      }
+    const items = get(bunsenModel, 'items.enum') || []
+
+    return items.map((value) => {
+      return { value }
     })
   },
 
   // == Functions ==============================================================
   /**
-   * This should be overriden by inherited inputs to convert the value to the appropriate format
+   * Convert value to appropriate format for template
    * @param {Boolean|String} data - value to parse
    * @returns {any} parsed value
    */
--- a/addon/components/inputs/checkbox-array.js
+++ b/addon/components/inputs/checkbox-array.js
@@ -1,0 +1,47 @@
+import AbstractInput from './abstract-input'
+import computed, {readOnly} from 'ember-computed-decorators'
+import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-checkbox-array'
+
+export default AbstractInput.extend({
+  // == Component Properties ===================================================
+
+  classNames: [
+    'frost-bunsen-input-checkbox-array',
+    'frost-field'
+  ],
+
+  layout,
+
+  // == State Properties =======================================================
+
+  selected: [],
+
+  // == Computed Properties ====================================================
+
+  @readOnly
+  @computed('bunsenModel')
+  options (bunsenModel) {
+    return bunsenModel.items.enum.map((key) => {
+      return {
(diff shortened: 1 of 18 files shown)


addon/components/inputs/checkbox-array.js
```diff
@@ -0,0 +1,47 @@
...
+  // == State Properties =======================================================
+
+  selected: [],
+
+  // == Computed Properties ====================================================
+
+  @readOnly
+  @computed('bunsenModel')
+  options (bunsenModel) {
+    return bunsenModel.items.enum.map((key) => {
```
sandersky: In the event the user omits `enum` in their model this will crash and burn. I'd make this a bit more robust with something like:

```js
import Ember from 'ember'
const {get} = Ember

…

  options (bunsenModel) {
    const items = get(bunsenModel, 'items.enum') || []

   return items.map((value) => {
      return {value}
    })
  },
```

addon/components/inputs/checkbox-array.js
```diff
@@ -0,0 +1,47 @@
...
+      return {
+        value: key
+      }
+    })
+  },
+
+  // == Functions ==============================================================
+  /**
+   * This should be overriden by inherited inputs to convert the value to the appropriate format
```
sandersky: This description only makes sense on the `AbstractInput`. Here I'd change it to say something like, "Convert value to appropriate format for template".

jsconfig.json
```diff
@@ -0,0 +1 @@
+{"compilerOptions":{"target":"es6","experimentalDecorators":true},"exclude":["node_modules","bower_components","tmp","vendor",".git","dist"]}
```
sandersky: This looks like a file that is probably for your editor and thus should be removed from your PR and added to `.gitignore`.

tests/dummy/app/pods/view/renderers/documentation/checkbox-array.md
```diff
@@ -0,0 +1,32 @@
+## checkbox-array
+
+This renderer provides a checkbox array input that allows for only mutiple options to be selected.
```
sandersky: Remove the word `only` from this line.

tests/dummy/app/pods/view/renderers/documentation/checkbox-array.md
```diff
@@ -0,0 +1,32 @@
...
+If the property type is *boolean* this input will render with the buttons `On` and `Off`, otherwise `enum` must be provided with a list of possible options.
```
sandersky: I believe this entire sentence can be removed.

addon/templates/components/frost-bunsen-input-checkbox-array.hbs
```diff
@@ -0,0 +1,22 @@
...
+  {{#if required}}
+    <div class='required'>Required</div>
+  {{/if}}
+</label>
+<div class={{inputWrapperClassName}}>
+  {{#each options as |option|}}
+    {{frost-checkbox
+      disabled=disabled
+      onInput=(action 'handleChange')
+      size=size
```
sandersky: In order for this to work you need to add the following to your component JavaScript.

```js
  @readOnly
  @computed('cellConfig')
  size (cellConfig) {
    return _.get(cellConfig, 'renderer.size') || 'medium'
  },
```

addon/components/inputs/checkbox-array.js
```diff
@@ -0,0 +1,47 @@
...
+   * @param {Boolean|String} data - value to parse
+   * @returns {any} parsed value
+   */
+  parseValue (data) {
+    var selected = this.get('selected')
```
sandersky: Not quite following this logic.

PoplarToppler: I am getting the reference to the **selected** state variable so that I didn't have to put this.get('selected') every time. Unless I'm crazy and the array isn't actually passing by reference, but it seemed to be. If you think that's not clear let me know and I'll just put this.get('selected'). Or is there a better way?

Unless you mean the whole block, in which case **data.value** is giving true/false if the checkbox is checked/unchecked respectively. So in the case that the checkbox is checked I push the _value_ of the checkbox (stored in **data.id**, coming from _value=option.value_ in the template) to the **selected** state array otherwise I remove it.

I'm fairly new to JavaScript as well, so if I'm doing something badly let me know.

sglanzer:
Deferring to @sandersky for the review, if you want to demo the functionality I can confirm the behavior, but I imagine you already know what it's supposed to be :)

tests/dummy/app/pods/view/renderers/documentation/checkbox-array.md
sandersky: @PoplarToppler I believe this comment is the only outstanding issue still.

PoplarToppler: Oops, not sure how I missed this one.

sandersky:
:+1:

PR #203, opened 2016-10-07
Fix required text on cells

### LINKED ISSUE #201: Required label incorrect
Given the following view and model there is required text under the label `Foo` which should not be present. This occurs when a nested property is required and one of it's ancestors is not. We need to make bunsen smarter in how it decides when to mark cells as required by taking into account ancestors of properties.

**Model**

```json
{
  "properties": {
    "foo": {
      "properties": {
        "bar": {
          "type": "string"
        }
      },
      "required": ["bar"],
      "type": "object"
    }
  },
  "type": "object"
}
```

**View**

```json
{
  "cells": [
    {
      "children": [
        {
          "model": "foo.bar"
        }
      ],
      "label": "Foo"
    }
  ],
  "type": "form",
  "version": "2.0"
}
```

screen shot 2016-10-07 at 7 01 07 am

sandersky:

Resolves #201 

# CHANGELOG

* **Fixed** issues with required label showing up on cells under scenarios where it shouldn't.

### DIFF
--- a/addon/components/cell.js
+++ b/addon/components/cell.js
@@ -108,16 +108,17 @@ export default Component.extend(PropTypeMixin, {
   },
 
   @readOnly
-  @computed('bunsenModel', 'bunsenView.cellDefinitions', 'cellConfig')
+  @computed('bunsenModel', 'bunsenView.cellDefinitions', 'cellConfig', 'value')
   /**
    * Determine whether or not cell contains required inputs
    * @param {BunsenModel} bunsenModel - bunsen model for form
    * @param {Object<String, BunsenCell>} cellDefinitions - list of cell definitions
    * @param {BunsenCell} mergedConfig - bunsen view cell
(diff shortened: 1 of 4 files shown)


cstolli:
👍

addon/components/cell.js
```diff
@@ -108,16 +108,17 @@ export default Component.extend(PropTypeMixin, {
   },
 
   @readOnly
-  @computed('bunsenModel', 'bunsenView.cellDefinitions', 'cellConfig')
+  @computed('bunsenModel', 'bunsenView.cellDefinitions', 'cellConfig', 'value')
```
sophypal: Might be safer to use `propagatedValue` here. Otherwise, `isRequired` would fire when it's descendant doesn't need to be re-rendered causing it's descendants to be re-rendered anyway. It might not even make a difference but one additional cell will be re-rendered.

PR #210, opened 2016-10-13
Better select

sophypal:

# CHANGELOG

* **Added** the option to specify static options for the `select` input
* **Added** the option to enable local filtering for the `select` input

### DIFF
--- a/addon/components/inputs/select.js
+++ b/addon/components/inputs/select.js
@@ -49,9 +49,10 @@ export default AbstractInput.extend({
   },
 
   @readOnly
-  @computed('bunsenModel', 'cellConfig.renderer.options')
-  listData (bunsenModel, renderOptions) {
+  @computed('bunsenModel', 'cellConfig')
+  listData (bunsenModel, cellConfig) {
     const enumDef = bunsenModel.items ? bunsenModel.items.enum : bunsenModel.enum
+    const renderOptions = _.get(cellConfig, 'renderer.options')
     const hasOverrides = !_.isEmpty(_.get(renderOptions, 'data'))
     const hasNoneOption = _.get(renderOptions, 'none.present')
 
@@ -78,9 +79,9 @@ export default AbstractInput.extend({
   },
 
   @readOnly
-  @computed('cellConfig.renderer.options.localFiltering')
-  isFilteringLocally (localFiltering = false) {
-    return localFiltering
+  @computed('cellConfig')
+  isFilteringLocally (cellConfig) {
+    return _.get(cellConfig, 'renderer.options.localFiltering') || false
   },
 
   // == Functions ==============================================================
--- a/addon/components/inputs/select.js
+++ b/addon/components/inputs/select.js
@@ -5,7 +5,7 @@ import _ from 'lodash'
 import Ember from 'ember'
 const {inject} = Ember
 import utils from 'bunsen-core/utils'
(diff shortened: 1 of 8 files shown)


addon/components/inputs/select.js
```diff
@@ -48,6 +48,41 @@ export default AbstractInput.extend({
...
+        value: ''
+      })
+      data = [none].concat(data)
+    }
+
+    return data
+  },
+
+  @readOnly
+  @computed('cellConfig.renderer.options.localFiltering')
```
sandersky: I've been trying to avoid using computed decorators on nested properties to make sure Ember doesn't convert our POJOs to Ember Objects. Instead I'd just do:

```js
@readOnly
@computed('cellConfig')
isFilteringLocally (cellConfig) {
  return _.get(cellConfig, 'renderer.options.localFiltering') || false
},
```

addon/templates/components/frost-bunsen-input-select.hbs
```diff
@@ -5,17 +5,31 @@
...
+      data=options
+      disabled=disabled
+      hook=hook
+      onChange=(action 'handleChange')
+      onFocusOut=(action 'showErrorMessage')
+      placeholder=cellConfig.placeholder
+      selectedValue=value
+    }}
+  {{else}}
+    {{frost-select
```
sandersky: It looks like the same code is in both the if and else conditions here. Can't you just completely revert the changes to this file?

tests/dummy/app/pods/view/renderers/documentation/select.md
```diff
@@ -22,3 +22,93 @@ This is the default renderer when the model contains the property `enum` or the
   "placeholder": "Select something…"
 }
 ```
+
+### Querying Data
```
sandersky: You mean we are finally going to document how all of this magic works? :+1:

addon/components/inputs/select.js
```diff
@@ -48,6 +48,41 @@ export default AbstractInput.extend({
     return false
   },
 
+  @readOnly
+  @computed('bunsenModel', 'cellConfig.renderer.options')
```
sandersky: I'd use `@computed('bunsenModel', 'cellConfig')` instead.

addon/templates/components/frost-bunsen-input-select.hbs
sophypal: One has `onInput` set and the other doesn't.

sandersky: I reread that 3 times and missed it every time. Thanks for clarifying.

sandersky: Ember really needs the spread operator for properties so we don't have to do this if/else duplication.

sandersky: Actually could you do the following?

```js
    {{frost-select
      class=valueClassName
      data=options
      disabled=disabled
      hook=hook
      onChange=(action 'handleChange')
      onFocusOut=(action 'showErrorMessage')
      onInput=(if isFilteringLocally undefined (action 'filterOptions'))
      placeholder=cellConfig.placeholder
      selectedValue=value
    }}
```

sandersky:
:+1:

sandersky:
:+1:

PR #212, opened 2016-10-13
Improve list-utils

sandersky:

# CHANGELOG

* **Fixed** list utility methods not to be less brittle and not throw an error when query isn't present.


sandersky:
:+1:

tests/dummy/mirage/data/models/wedding-application.js
```diff
@@ -5,7 +5,6 @@ export default {
         countryOfBirth: {
           labelAttribute: 'name',
           modelType: 'country',
-          query: {},
```
sophypal: Confused about this change. Isn't having a non-undefined `query` object the signal to fetch this from the store?

PR #214, opened 2016-10-14
Fix query

sophypal:

# CHANGELOG

* **Fixed** issue where the `select` input's `query` isn't specified and we should have been looking at `modelType` to fetch from the store.


sophypal:
👍

PR #215, opened 2016-10-14
Fix query

sophypal:

# CHANGELOG

* **Fixed** issue with `hasQueryChanged` returning false when form values actually differed when the old value had missing dependencies.


sophypal:
👍

PR #219, opened 2016-10-14
Fix shift when field goes into error

sandersky:

# CHANGELOG

* **Fixed** input position shifting down 3 pixels when field goes into error state.


sandersky:
:+1:

PR #224, opened 2016-10-18
Fix issue where query params had to be strings

job13er:

# CHANGELOG

* **Fixed** `list-utils` module to no longer assume that all query params are string values. We were blindly calling `.replace()` to swap out `$filter` if present. Now, we make sure it's actually a string before trying to replace something in it.

### DIFF
--- a/.bithoundrc
+++ b/.bithoundrc
@@ -1,6 +1,7 @@
 {
   "dependencies": {
     "mute": [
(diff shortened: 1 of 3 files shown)


job13er:
@TheOtherDude just for you :)

sandersky:
:+1:

sandersky:
:+1:

PR #231, opened 2016-10-20
Add more select tests

### LINKED ISSUE #229: Encountering a problem with enums not having a value when pre-populated
When creating an enum value and pre-populating it, we noticed that if the value was not changed, bunsen reported `undefined` for that field. We need tests around this and to make sure this works as expected.

sandersky:

Resolves #229 

# CHANGELOG

* **Added** more tests around enum driven select to make sure it plays nice with initial values and default values.


sandersky:
:+1:

PR #242, opened 2016-10-24
Make link renderer more configurable

sandersky:

# CHANGELOG
* **Added**`route` and `url` options to `link` renderer.

### DIFF
--- a/addon/components/inputs/link.js
+++ b/addon/components/inputs/link.js
@@ -15,10 +15,21 @@ export default AbstractInput.extend({
 
   // == Computed Properties ====================================================
 
-  // We totally don't care about this cause it's view schema
   @readOnly
   @computed('cellConfig', 'value')
   linkLabel (cellConfig, value) {
     return _.get(cellConfig, 'renderer.label') || value
+  },
+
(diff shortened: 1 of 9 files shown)


addon/components/inputs/link.js
```diff
@@ -15,10 +16,63 @@ export default AbstractInput.extend({
...
+    const rendererLabel = _.get(cellConfig, 'renderer.label')
+    const rendererUrl = _.get(cellConfig, 'renderer.url')
+    const labelContainsReferences = rendererLabel && rendererLabel.indexOf('${') !== -1
+    const urlContainsReferences = rendererUrl && rendererUrl.indexOf('${') !== -1
+
+    if (!labelContainsReferences && !urlContainsReferences) {
+      return
+    }
+
+    this.set('formValue', newValue)
```
sophypal: Since `formValue` is used in the above computed property, do you think it'll be worthwhile to diff against the changes to the label and url references. Otherwise, it'll cause a re-render for every form value change but not a big deal unless you expect to see tons of these on a single form.

sophypal:
👍

PR #254, opened 2016-10-27
Make sure to only select a tab if it is still in the list

sandersky:

# CHANGELOG

* **Fixed** bug where tabs would all become unselected when entire bunsen state changed.


sandersky:
:+1:

PR #261, opened 2016-11-04
DO NOT MERGE – checkbox-array changes

PoplarToppler:

# CHANGELOG

* **Fixed** an issue where checkbox selection was carrying across checkbox-array groups
* **Fixed** checkbox-array styling, checkboxes are now in a column
* **Added** spacing between checkboxes for object-browser facets styles
* **Added** renderer.labels for overriding labels
* **Added** Style for frost-object-browser-facets for checkbox array spacing
* **Updated** `bunsen-core` to latest version.
* **Updated** `ember-bunsen-core` to latest version.


addon/components/inputs/checkbox-array.js
```diff
@@ -5,6 +5,14 @@ import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-c
 import Ember from 'ember'
 const {get} = Ember
 
+export const helpers = {
+  validateChoices (choices, meta) {
+    if (!Ember.isEmpty(choices) && !Ember.isEmpty(meta)) {
```
sandersky: Change line 6 to be `const {get, isEmpty} = Ember` and just use `isEmpty` in this line.

addon/components/inputs/checkbox-array.js
```diff
@@ -27,13 +35,36 @@ export default AbstractInput.extend({
   },
 
   @readOnly
-  @computed('bunsenModel')
-  options (bunsenModel) {
+  @computed('bunsenModel', 'cellConfig')
+  options (bunsenModel, cellConfig) {
     const items = get(bunsenModel, 'items.enum') || []
+    const choices = _.get(cellConfig, 'renderer.choices') || []
```
sandersky: Is there a reason you use Lodash's get method instead of Ember's?

addon/components/inputs/checkbox-array.js
```diff
@@ -27,13 +35,36 @@ export default AbstractInput.extend({
...
+    const meta = _.get(cellConfig, 'renderer.meta') || []
+
+    helpers.validateChoices(choices, meta)
 
-    return items.map((value) => {
-      return { value }
-    })
+    // Initialize selected for each model property
+    this.get('selected')[this.get('bunsenId')] = []
```
sandersky: This change is the wrong approach. You have a `selected` property set on the class itself which is being shared across all instances. The proper way to do this would instead be:

```js
const assign = Object.assign || Ember.assign || Ember.merge
…
  getDefaultProps () {
    const defaults = this._super(...arguments)
    return assign(defaults, {
      selected: []
    })
  }
```

This will give each instance it's own value for `selected` and you can maintain the array approach.

addon/components/inputs/checkbox-array.js
```diff
@@ -17,7 +27,12 @@ export default AbstractInput.extend({
 
   // == State Properties =======================================================
 
-  selected: [],
+  getDefaultProps () {
```
sglanzer: @sandersky I assume we're not on prop-types 3.0?

tests/dummy/app/pods/view/renderers/models/checkbox-array.js
```diff
@@ -6,6 +6,19 @@ export default {
         type: 'string'
       },
       type: 'array'
+    },
+    serviceAffecting: {
+      items: {
+        type: 'string'
```
job13er: I'm pretty sure this also needs an `enum: ['SERVICE_AFFECTING', 'NOT_SERVICE_AFFECTING']` to be accurate.

tests/dummy/app/pods/view/renderers/models/checkbox-array.js
```diff
@@ -6,6 +6,19 @@ export default {
...
+      },
+      type: 'array'
```
job13er: Unless you want `['SERVICE_AFFECTING', 'NOT_SERVICE_AFFECTING', 'SERVICE_AFFECTING']` to be valid as well, you probably also want `uniqueItems: true` here

 https://spacetelescope.github.io/understanding-json-schema/reference/array.html#uniqueness

tests/dummy/app/pods/view/renderers/documentation/checkbox-array.md
```diff
@@ -29,3 +29,36 @@ Change what size checkboxes are used. See [ember-frost-core](http://ciena-frost.
...
+#### renderer.selectedValues
+
+Allows checkboxes to be pre-checked. E.g. for retaining checks from query params upon refresh.
+
+```json
+{
+  "model": "foo",
+  "renderer": {
+    "name": "checkbox-array",
+    "selectedValues": ["values", "of", "selected", "checkboxes"]
```
job13er: I don't think you need a separate `selectedValues` in the view, can't this be driven by the current value of that part of the model? Or by `default` specified in the model itself?

tests/dummy/app/pods/view/renderers/views/checkbox-array.js
```diff
@@ -3,12 +3,54 @@ export default {
...
+          label: 'Severity',
+          collapsible: true
+        },
+        {
+          children: [
+            {
+              model: 'serviceAffecting',
+              renderer: {
+                name: 'checkbox-array',
+                data: [
```
job13er: Rather than this being an array of key/value pairs, I'd suggest a hash that maps the already defined `enum` values from the model to user-friendly labels:

```js
labels: {
  'NOT_SERVICE_AFFECTING': 'No',
  'SERVICE_AFFECTING': 'Yes'
}
```

sandersky:
I think we have three use cases to solve with this renderer and the API should look like the following in my opinion.

1. We know the list of options up-front and the labels match the values.

    **Model**

    ```json
    {
      "properties": {
        "alarms": {
          "items": {
            "enum": ["Critical", "Major", "Minor", "Warning"],
            "type": "string"
          },
          "type": "array",
          "uniqueItems": true
        }
      },
      "type": "object"
    }
    ```

    **View**

    ```json
    {
      "cells": [
        {
          "model": "alarms",
          "renderer": {
            "name": "checkbox-array"
          }
        }
      ],
      "type": "form",
      "version": "2.0"
    }
    ```

2. We know the list of options up-front but the labels don't match the values.

    **Model**

    ```json
    {
      "properties": {
        "alarms": {
          "items": {
            "enum": ["CRITICAL", "MAJOR", "MINOR", "WARN"],
            "type": "string"
          },
          "type": "array",
          "uniqueItems": true
        }
      },
      "type": "object"
    }
    ```

    **View**

    ```json
    {
      "cells": [
        {
          "model": "alarms",
          "renderer": {
            "name": "checkbox-array",
            "labels": {
              "CRITICAL": "Critical",
              "MAJOR": "Major",
              "MINOR": "Minor",
              "WARN": "Wanring"
            }
          }
        }
      ],
      "type": "form",
      "version": "2.0"
    }
    ```

3. We want to data-drive the list of options from an API.

    **Model**

    ```json
    {
      "properties": {
        "alarms": {
          "items": {
            "labelAttribute": "name",
            "modelType": "alarm-type",
            "query": {
              "filterEmpty": true
            },
            "type": "string",
            "valueAttribute": "id"
          },
          "type": "array",
          "uniqueItems": true
        }
      },
      "type": "object"
    }
    ```

    **View**

    ```json
    {
      "cells": [
        {
          "model": "alarms",
          "renderer": {
            "name": "checkbox-array"
          }
        }
      ],
      "type": "form",
      "version": "2.0"
    }
    ```

    > Note: in this case the `modelType`, `labelAttribute`, `valueAttribute`, and `query` parameters should be able to be defined in the model or view which is how the `select` renderer works.

tests/dummy/app/pods/view/renderers/documentation/checkbox-array.md
PoplarToppler: I'm not sure what you mean by "current value of that part of the model". 

I feel like this is pretty straight forward and simple to use and can be easily set by query params. See https://bitbucket.ciena.com/users/cc40498/repos/ciena-alarms-engine/browse/app/pods/alarms/modules/alarms-serializer.js#125-132 for the use case I am trying to provide. It's easy to simply take the query params and put them in there.

Are there a reasons why we wouldn't want it this way?
Don't get me wrong, if it is bad I am fine with changing it.

tests/dummy/app/pods/view/renderers/views/checkbox-array.js
PoplarToppler: I had done it this way originally. @sandersky had said he thought it was overly complicated and preferred to simply override the enum completely. But if you both agree we can look at it agian.

PoplarToppler:
I agree with the first two use cases. We can provide functionality for the third one later; but I don't think we want to use that type of data drive, at least in the object browser. 

Steven and Joel and I discussed this and came to the conclusion that we don't want each type of facet block in the object browser to be making separate REST calls. We want all the facets to be populated via meta data from the original context request into the browser (e.g. keyword search + date/time ranges for alarms). @sglanzer can correct me if I'm wrong, maybe this needs to be discussed more.

Assuming that we don't want separate REST calls I think you're missing a fourth use case where neither the labels nor the options are known up front and come completely from the meta data of the original context request (e.g. keyword search and date/time ranges for alarms currently). 

You could argue that at some level we must know the complete set of options up front in most (maybe all) cases. Then we could have the complete set in in the model to begin with and decide which ones to display based upon which labels are provided in something similar to what @job13er mentioned regarding the data property.

sandersky:
@PoplarToppler This 4th use case would really be either use case 1 or 2 driven by computed properties. For example:

```js

alarmTypes: {
  critical: 'Critical',
  major: 'Major',
  minor: 'Minor',
  warning: 'Warning'
},

@readOnly
@computed('alarmTypes')
bunsenModel (alarmTypes) {
  return {
    properties: {
      items: {
        enum: Object.keys(alarmTypes),
        type: 'string'
      },
      type: 'array',
      uniqueItems: true
    },
    type: 'object'
  }
},

@readOnly
@computed('alarmTypes')
bunsenView (alarmTypes) {
  return {
    cells: [
      {
        model: 'alarms',
        renderer: {
          name: 'checkbox-array',
         labels: alarmTypes
        }
      }
    ],
    type: 'form',
    version: '2.0'
  }
}
```

sandersky:
I want to drive home one key concept here: With bunsen the view is not intended to drive the data but rather how the data is displayed. The model is responsible for determining what the data is. In the case of label mappings you aren't changing the list of items but rather what text is shown for each item.

tests/dummy/app/pods/view/renderers/models/checkbox-array.js
job13er: That's kind of my point, you shouldn't override possible values in the view, unless they're factored into the validation.

tests/dummy/app/pods/view/renderers/documentation/checkbox-array.md
job13er: I mean the `bunsenValue` you pass into the component, it will have a `foo` property that can have a  value of `['values', 'of', 'selected', 'checkboxes']` just like every other input in a bunsen form. I'm not sure why you want to use some other mechanism for setting the value of something, other than the `value` property.

PoplarToppler:
Thanks for your comments guys!

PoplarToppler:
Oops I pressed the wrong button.

addon/components/inputs/checkbox-array.js
```diff
@@ -27,13 +31,23 @@ export default AbstractInput.extend({
...
+    if (isEmpty(labels)) {
+      options = items.map((item) => {
+        return { value: item, label: item, checked: value.indexOf(item) > -1 }
+      })
+    } else {
+      options = items.map((item) => {
+        return { value: item, label: labels[item] || item, checked: value.indexOf(item) > -1 }
+      })
+    }
+    return options
```
sandersky:
```js
  @readOnly
  @computed('bunsenModel', 'cellConfig', 'value')
  options (bunsenModel, cellConfig, value) {
    const items = get(bunsenModel, 'items.enum') || []
    const labels = get(cellConfig, 'renderer.labels') || {}

    value = value || []

    return items.map((item) => {
      return {
        checked: value.indexOf(item) > -1,
        label: labels[item] || item,
        value: item
      }
    })
  },
```

sandersky:
:+1:

PR #268, opened 2016-11-18
Fixed bug in multi-select renderer

TheOtherDude:

# CHANGELOG

* **Fixed** a bug in multi-select renderer that occurred when the value was set to an array the same length as the existing value


sandersky:
:+1:

PR #271, opened 2016-11-29
Fix required labels

### LINKED ISSUE #270: Section showing up as required when it shouldn't be
*Model*

```json
{
  "properties": {
    "foo": {
      "properties": {
        "bar": {"type": "string"},
        "baz": {"type": "string"}
      },
        "required": ["bar", "baz"],
      "type": "object"
    }
  },
  "type": "object"
}
```

*View*

```json
{
  "cells": [
    {
      "label": "Test",
      "model": "foo",
      "children": [
        {"model": "bar"},
        {"model": "baz"}
      ]
    }
  ],
  "type": "form",
  "version": "2.0"
}
```

*Value*

```json
{}
```

screen shot 2016-11-28 at 2 34 58 pm

When `foo` isn't present the section containing `foo.bar` and `foo.baz` should not be marked as required because the two fields are only required when `foo` is present, which is not a required property.

sandersky:

resolves #270 

# CHANGELOG

* **Added** missing tests for password renderer to reach full coverage of renderer.
* **Fixed** bug where sections were sometimes showing required label when they shouldn't.


cstolli:
👍

PR #290, opened 2016-12-07
BPSO 33985 - open tab using tabName

sushant1245:

# CHANGELOG

* **Updated** `frost-bunsen-detail` to support a new property, `selectedTabLabel`, to allow a consumer to set which tab is currently in view.


addon/components/detail.js
```diff
@@ -101,6 +101,7 @@ export default Component.extend(PropTypeMixin, {
     ]),
     hook: PropTypes.string,
     onChange: PropTypes.func,
+    tabName: PropTypes.string,
```
sandersky: This property name isn't clear as to what it is doing. Can we rename this to be `selectedTabLabel`.

sandersky: Also please put the property in the correct location in the `propTypes` object to maintain alphabetical sorting.

addon/components/detail.js
```diff
@@ -384,6 +384,17 @@ export default Component.extend(PropTypeMixin, {
     this.set('reduxStore', reduxStore)
     this.set('renderModel', reduxStore.getState().model)
     reduxStore.subscribe(this.storeUpdated.bind(this))
+
+    /**
```
sandersky: Please remove this comment block as it adds no value.

addon/components/detail.js
```diff
@@ -384,6 +384,17 @@ export default Component.extend(PropTypeMixin, {
...
+     * when tabName is send while calling component
+     */
+    let tabName = this.get('tabName')
+    if (tabName !== undefined) {
+      const selectedTab = this.get('cellTabs').findBy('alias', tabName)
+      if (selectedTab !== undefined) {
+        this.set('selectedTabIndex', selectedTab.id)
+      }
+    }
```
sandersky: This logic shouldn't go in the init method because a consumer may programmatically replace the entire bunsen model/view on an instance after it is initialized, at which point they'd expect the correct tab to be selected. Instead this logic should go in `didReceiveAttrs` and execute anytime the value changes.

tests/unit/components/detail-test.js
```diff
@@ -464,6 +464,58 @@ describeComponent(...unitTest('frost-bunsen-detail'), function () {
...
+        sandbox.stub(component, 'setProperties')
+        component.init()
+      })
+
+      it('check if selectedTabIndex start with 1', function () {
+        const selectedTabIndex = component.get('selectedTabIndex')
+        expect(selectedTabIndex.substr(0, selectedTabIndex.indexOf('-'))).to.equal('1')
+      })
+    })
+
```
sandersky: Please remove this unit test and instead add an integration test at `tests/integration/frost-bunsen-detail/selected-tab-name-test.js`.

sandersky:
@sushant1245 also I updated your PR description to keep the checkboxes (you don't need to replace them just select one). In the case of this PR you are adding a feature rather than fixing a bug so it should be a minor change (updated that in the description as well).

sandersky:
:-1:

sandersky:
@psbanka this failed as it needed merged with upstream. I guess when other PR's are merged it doesn't force Travis to rerun CI on previously run and still open PR's. https://travis-ci.org/ciena-frost/ember-frost-bunsen/builds/183499455

I'm going to reset `master` to the last commit before this PR was merged so the entire repo isn't in a bad state. Can you have @sushant1245 create a new PR with these changes merged with upstream?

PR #298, opened 2016-12-09
Add new frost-bunsen component and add spread as well as a few bug fixes

sandersky:

# CHANGELOG

* **Added** new component `frost-bunsen` which is designed to be a single entry point for rendering both forms and detail views, as well as any future view types.
* **Added** [spread](https://github.com/ciena-blueplanet/ember-spread) operator to existing `frost-bunsen-detail` and `frost-bunsen-form` components.
* **Fixed** `geolocation` renderer to work when consumer latitude and/or longitude properties expect the format to be a number instead of a string.
* **Fixed** description bubble property validations to not require `description` property as it was causing undesired `ember-prop-type` warnings in the console.
* **Fixed** issue where enter key would sometimes submit form by setting `onsubmit='return false'` on the form element itself.
* **Fixed** issue where `frost-bunsen-form` and `frost-bunsen-detail` would try to update state after the component had been destroyed when getting an update from the redux store.

### DIFF
--- a/addon/components/description-bubble.js
+++ b/addon/components/description-bubble.js
@@ -12,6 +12,6 @@ export default Component.extend(PropTypesMixin, {
   // == State Properties =======================================================
 
   propTypes: {
-    description: PropTypes.string.isRequired
+    description: PropTypes.string
   }
 })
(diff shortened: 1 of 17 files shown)


sandersky:
:+1:

sglanzer:
🎉

PR #309, opened 2016-12-13
Next Major Release

sandersky:

# CHANGELOG

* **Cleaned** up a bunch of tests by replacing boilerplate with test helpers.

  > Note: These test helpers will get wired into a consumers test suite but are still considered private API. If you choose to start using them know they may change in breaking ways before the next major on this project happens.

* **Removed** `hide` and `show` icons which are no longer used. If you were using these icons in your app via something like below you'll no longer have access to them:

  ```handlebars
  {{frost-icon icon='show' pack='frost-bunsen'}}
  ```

* **Replaced** `ember-redux` with `ember-redux-shim` and `ember-redux-thunk`.

  > NOTE: When upgrading you can remove `ember-browserify` and `ember-redux` from your `package.json` if you are not using them in your app and no other addons you consume require them.


sandersky:
:+1:

PR #310, opened 2016-12-14
Bpso 33985

sushant1245:

# CHANGELOG

BPSO-33985 be able to see the provisioning log when the Error Log link is clicked in the list view
open a specific tab in tab-nav


tests/integration/components/frost-bunsen-detail/selected-tab-name-test.js
```diff
@@ -0,0 +1,61 @@
+import {expect} from 'chai'
```
sandersky: I've added new test helpers I'd like to see you use here as well as a test to verify you can update the `selectedTabLabel` property to programmatically change the tab.

```js
import {expect} from 'chai'
import {setupFormComponentTest} from 'dummy/tests/helpers/utils'
import hbs from 'htmlbars-inline-precompile'
import {beforeEach, describe, it} from 'mocha'

describe('Integration: Component / frost-bunsen-detail / open specific tab using tab label', function () {
  setupFormComponentTest({
    bunsenModel: {
      properties: {
        bar: {type: 'number'},
        baz: {type: 'boolean'},
        foo: {type: 'string'}
      },
      type: 'object'
    },
    bunsenView: {
      cellDefinitions: {
        one: {
          children: [
             {model: 'foo'},
             {model: 'bar'}
          ]
        },
        two: {
          children: [
             {model: 'baz'}
          ]
        }
      },
      cells: [
         {label: 'One', extends: 'one'},
         {label: 'Two', extends: 'two'}
      ],
      type: 'form',
      version: '2.0'
    },
    selectedTabLabel: 'Two'
  })

  it('renders second tab', function () {
    expect(this.$('.frost-tabs .active div').text().trim()).to.equal('Two')
  })

  describe('when selectedTab property updated to be first tab', function () {
    beforeEach(function () {
      this.set('selectedTabLabel', 'One')
    })

    it('renders first tab', function () {
      expect(this.$('.frost-tabs .active div').text().trim()).to.equal('One')
    })
  })
})
```

sandersky:
@sushant1245 I think we will also need a new callback property `onTabChange` on the detail component that informs the consumer whenever the tab changes. Otherwise we risk a consumer that passes in the initial tab accidentally causing the tab to switch back to the initial tab when a user has manually selected another tab.

tests/integration/components/frost-bunsen-detail/selected-tab-name-test.js
sandersky: My point is that you render a bunsen detail view with the selectedTabLabel as `Two`. The UI comes up and the second tab is selected. You then programmatically change the selectedTabLabel property to `One`. At this point I'd expect the first tab to become selected. This test isn't testing your use case in particular but bunsen is an open source addon consumed by many people with different use cases. If we are going to add this feature I want to make sure it is versatile in it's usage for whatever use cases come up.

sushant1245: yes i agree , now it make sense

sushant1245:
I will submit changes suggested by @sandersky in upcoming PR

PR #313, opened 2016-12-15
Fix issue where checkboxes have left padding

sandersky:

# CHANGELOG

* **Fixed** bug where `frost-link` class was being applied to all inputs instead of just link input which was causing checkboxes to get an undesired left padding.


sandersky:
:+1:

PR #317, opened 2016-12-16
Make solution more robust

sandersky:

# CHANGELOG

* **Fixed** CSS fix from previous release to be more robust as it only worked when `ember-frost-core`'s SASS wasn't imported after `ember-frost-bunsen`'s SASS.


sandersky:
:+1:

PR #321, opened 2016-12-19
12.x

sandersky:

# CHANGELOG

* **Added** new `onTabChange` property so consumers can be notified of changing tabs.
* **Added** new `selectedTabLabel` property so consumers can set which tab is selected/visible.


sandersky:
:+1:

PR #330, opened 2017-01-18
Update dependencies and some minor cleanup

sandersky:

# CHANGELOG

* Cleaned up some code in preparation to get this addon working with Ember 2.10.
* **Updated** dependencies to latest versions (shims were renamed which is why is a minor and not a major). If a consumer uses the old shims instead it'll still function the same.


sandersky:
:+1:

PR #333, opened 2017-01-20
Get codebase working with Ember 2.10

sandersky:

# CHANGELOG

* **Fixed** issues in codebase that were preventing it from being used with Ember 2.10. One known issue currently is that when using with Ember 2.10 the `autofocus` property for forms won't function. If using an older version of Ember the `autofocus` property will continue to work as it always has.


sophypal:
👍

PR #334, opened 2017-01-20
DO NOT MERGE – Update formValueChanged to support tasks

sandersky:

# CHANGELOG

* **Added** support to use `ember-concurrency` tasks for `formValueChanged` property so you don't have to do the additional `this.isDestroyed || this.isDestroying` check.

### DIFF
--- a/addon/components/detail.js
+++ b/addon/components/detail.js
@@ -363,7 +363,11 @@ export default Component.extend(SpreadMixin, HookMixin, PropTypeMixin, {
 
       this.get('registeredComponents').forEach((component) => {
         if (!component.isDestroyed) {
-          component.formValueChanged(value)
+          if (typeOf(component.formValueChanged) === 'function') {
+            component.formValueChanged(value)
+          } else {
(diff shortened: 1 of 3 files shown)


addon/components/detail.js
```diff
@@ -588,8 +592,14 @@ export default Component.extend(SpreadMixin, HookMixin, PropTypeMixin, {
         component.unregisterForFormValueChanges(component)
       })
 
+      const formValue = this.get('renderValue') || {}
+
       // Make sure we inform component of formValue immediately
-      component.formValueChanged(this.get('renderValue') || {})
+      if (typeOf(component.formValueChanged) === 'function') {
```
cstolli: is the northbound interface here guaranteed to either pass a `function` or a `task` here?

cstolli: shouldn't it be `get('formValueChanged')` either way?  Or is the intent here to respect a hardwired `formValueChange` method on the component?  What if `formValueChanged` is passed, but as a function, not a task?

PR #339, opened 2017-01-26
Added date renderer based on ember-frost-date-picker

TheOtherDude:

Rsolves #126

# CHANGELOG

**Added**  new `date` renderer for handling date input.


tests/integration/components/frost-bunsen-form/renderers/date-test.js
```diff
@@ -0,0 +1,461 @@
...
+        label: 'Foo'
+      })
+      expectOnChangeState(ctx, {foo: '2017-01-24'})
+      expectOnValidationState(ctx, {count: 1})
+    })
+
+    describe('when date is unset', function () {
+      beforeEach(function () {
+        ctx.props.onValidation.reset()
+        this.set('value', {})
```
TheOtherDude: As far as I can tell, it's currently impossible for the user to clear the value of an `ember-frost-date-picker`, so clearing it programmatically will have to do?

https://github.com/ciena-frost/ember-frost-date-picker/issues/26

sandersky: The same is still true for a `frost-select` as well. Not really bunsen's problem per say. I think the issue belongs to the `frost-date-picker` itself, once they support it bunsen should get it for free.

addon/components/inputs/date.js
```diff
@@ -0,0 +1,28 @@
...
+  layout,
+
+  // == Computed Properties ====================================================
+
+  // == Functions ==============================================================
+
+  // == Actions ===============================================================
+
+  actions: {
+    selectDate (pikaday) {
```
sandersky: Add JSDoc please, I have no idea what the format of `pikaday` is. String? Object? If so, waht keys/values?

app/styles/ember-frost-bunsen/frost-date-picker.scss
```diff
@@ -0,0 +1,8 @@
+.frost-bunsen-has-value {
```
TheOtherDude: I don't know if this styling should go in `ember-frost-date-picker` or not. The default text it uses looks like a placeholder (to me at least!). We've been using this in quote.

addon/templates/components/frost-bunsen-input-date.hbs
```diff
@@ -0,0 +1,22 @@
...
+    onFocusIn=(action 'hideErrorMessage')
+    onFocusOut=(action 'showErrorMessage')
+    onSelect=(action 'selectDate')
+  }}
+</div>
+{{#if renderErrorMessage}}
+  <div class="frost-bunsen-error">
+    {{renderErrorMessage}}
+  </div>
+{{/if}}
```
sandersky: I know it is minor but can you add a newline to the end of the file (you can probably configure your editor to always do this for you on save).

bower.json
```diff
@@ -10,6 +10,9 @@
     "pretender": "1.4.2",
     "prism": "1.6.0",
     "showdown": "1.5.1",
-    "sinonjs": "1.17.1"
+    "sinonjs": "1.17.1",
+    "moment": "^2.17.1",
```
sandersky: Why is moment here? We already include moment via npm afaik.

bower.json
```diff
@@ -10,6 +10,9 @@
...
+    "pikaday": "~1.5.1",
+    "clockpicker-seconds": "~0.1.6"
```
sandersky: I'm strongly against bringing in more bower dependencies. I've been trying to eliminate them for a while as has Ember itself. IMO these really need to be shimmed so we can include them via NPM instead, I'll even help make that happen.

app/styles/ember-frost-bunsen/frost-date-picker.scss
sandersky: Yeah don't add it here. If the default looks like garbage the correct thing todo is add the improved styling to the `ember-frost-date-picker` addon itself. Otherwise we'll have inconsistencies between apps using/not using bunsen.

TheOtherDude:
There's actually an issue with removing the date picker CSS.

The date picker uses moment to select today's date if a `currentValue` is not passed to it. However, it does not call `onSelect` with that default value, so Bunsen doesn't get it. 

The easy fix would be to make `ember-frost-date-picker` call `onSelect` when it sets a default value, but we actually need to support unset dates (eg for optional form values). 

Our workaround in quote, where the default value (that Bunsen didn't get) looked like a placeholder and then setting a 'real' value made the text darker actually worked pretty well...

addon/components/inputs/date.js
```diff
@@ -0,0 +1,31 @@
...
+  // == Functions ==============================================================
+
+  // == Actions ===============================================================
+
+  actions: {
+    /**
+     * Send the new date to handleChange()
+     * @param {String} dateString - the selected date, formatted as YYYY-MM-DD
+     */
+    selectDate (dateString) {
```
sandersky: It looks like this method doesn't do anything special. Why not remove it and just have your template use the `handleChange` action.

sandersky:
:+1:

PR #342, opened 2017-01-27
fix checkbox-array bug where selected is not kept on refresh

PoplarToppler:

# CHANGELOG

**Fixed** a checkbox-array bug where keeping the value upon refresh was not kept in the selected property, causing the checkboxes to all be cleared upon checking any of the checkboxes after a refresh.

### DIFF
--- a/addon/components/inputs/checkbox-array.js
+++ b/addon/components/inputs/checkbox-array.js
@@ -18,12 +18,6 @@ export default AbstractInput.extend({
 
   // == State Properties =======================================================
 
-  getDefaultProps () {
-    return {
-      selected: _.clone(this.get('value')) || []
-    }
-  },
-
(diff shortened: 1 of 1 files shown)


addon/components/inputs/checkbox-array.js
```diff
@@ -20,7 +20,7 @@ export default AbstractInput.extend({
 
   getDefaultProps () {
     return {
-      selected: []
+      selected: _.clone(this.get('value')) || []
```
sandersky: Instead can you do this:
```js
selected: Array.from(this.get('value') || [])
```

sandersky:
Also please add a test that exercises the bug so the fix doesn't regress in the future.

sandersky:
:+1:

PR #346, opened 2017-02-01
Added queryForCurrentValue option to select renderer

TheOtherDude:

# CHANGELOG

**Added** queryForCurrentValue option to select renderer


addon/list-utils.js
```diff
@@ -142,13 +144,30 @@ export function getItemsFromEmberData (value, modelDef, data, bunsenId, store, f
...
+  })
+
+  return queries.then((queries) => {
+    const {items, valueRecord} = queries
+    if (actuallyFindCurrentValue &&
+      shouldAddCurrentValue({data: items, valueRecord, labelAttribute, valueAttribute, filter})) {
+      return normalizeItems({data: items, labelAttribute, records: [valueRecord], valueAttribute})
+    }
+    return items
+  })
```
sandersky: You could just do:

```js
return RSVP.hash({
  …
})
  .then(({items, valueRecord}) => {
    …
  })
```

addon/list-utils.js
```diff
@@ -176,3 +195,19 @@ function normalizeItems ({data, labelAttribute, records, valueAttribute}) {
...
+ * @param {String} labelAttribute - dot notated path to label attribute in record
+ * @param {String} valueAttribute - dot notated path to value attribute in record
+ * @param {String} filter - the partial match query filter to populate
+ * @returns {boolean} true if valueRecord should be added to data
+ */
+function shouldAddCurrentValue ({data, valueRecord, labelAttribute, valueAttribute, filter}) {
+  const filterRegex = new RegExp(filter, 'i')
+  return filterRegex.test(valueRecord.get(labelAttribute)) && !data.find(item => {
+    return item.value === valueRecord.get(valueAttribute)
+  })
```
sandersky: I'd probably do something like this so it is more clear to someone else exactly what this is doing:

```js
function shouldAddCurrentValue ({data, filter, labelAttribute, valueAttribute, valueRecord}) {
  const filterRegex = new RegExp(filter, 'i')
  const selectedItemMatchesFilter = filterRegex.test(valueRecord.get(labelAttribute))
  const itemsContainSelectedItem = data.find(item => item.value === valueRecord.get(valueAttribute))
  return selectedItemMatchesFilter && !itemsContainSelectedItem
}
```

addon/list-utils.js
```diff
@@ -133,7 +133,9 @@ export function getItemsFromAjaxCall ({ajax, bunsenId, data, filter, options, va
  */
 export function getItemsFromEmberData (value, modelDef, data, bunsenId, store, filter) {
   const modelType = modelDef.modelType || 'resources'
-  const {labelAttribute, valueAttribute} = modelDef
+  const {labelAttribute, queryForCurrentValue, valueAttribute} = modelDef
+  const valueAsId = Number(value[bunsenId])
+  const actuallyFindCurrentValue = queryForCurrentValue && !!valueAsId
```
sandersky: I'm not sure I see the need to have `queryForCurrentValue`. Is there a reason not to just always query for the current value? If you can think of a use case in which it'd break then maybe make it opt-out instead of opt-in?

TheOtherDude: What if someone has a number value that isn't an id? 

And if you don't have a paged API, why not avoid the extra query?

sandersky: Fair enough, I'm convinced.

sandersky:
:+1:

PR #349, opened 2017-02-03
Updated code for datetime renderer to hold data correctly.

ZarifS:

# CHANGELOG

Updated the datetime renderer to allow for correct data storage for the bunsen forms.

### DIFF
--- a/addon/components/inputs/datetime.js
+++ b/addon/components/inputs/datetime.js
@@ -1,53 +1,55 @@
 import AbstractInput from './abstract-input'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-datetime'
+import Ember from 'ember'
+import { Format, setTime, validateTime } from 'ember-frost-date-picker'
+import moment from 'moment'
 
 export default AbstractInput.extend({
 
-  // == Component Properties ===================================================
+    // == Component Properties ===================================================
 
-  classNameBindings: ['value:frost-bunsen-has-value'],
+    classNameBindings: ['value:frost-bunsen-has-value'],
 
-  classNames: [
-    'frost-bunsen-input-datetime',
-    'frost-field'
-  ],
+    classNames: [
+        'frost-bunsen-input-datetime',
+        'frost-field'
+    ],
 
-  layout,
+    layout,
 
-  // == State Properties =======================================================
+    dateTimeValue: moment().subtract(1, 'day').subtract(1, 'hour').format(Format.dateTime),
+    dateTimeValueInvalid: false,
 
-  getDefaultProps () {
-    if (this.get('value')) {
-      var arr = this.get('value').split('T')
+    // == Actions ===============================================================
 
-      return {
-        date: arr[0] || '',
-        time: arr[1] || ''
-      }
+    actions: {
+
+        onDateTimeChange(dateTimeValue) {
+
+            const momentDateTimeValue = moment(dateTimeValue)
+            if (!momentDateTimeValue.isValid()) {
+                this.set('dateTimeValueInvalid', true)
+                return 
+            }
+
+            //Comment this  else statment out if using with validation. 
+            else {
+                this.send('handleChange', dateTimeValue)
+            }
+
+            // Example just to show validation - any date after 2015-01-01T10:00:00-00:00 is valid. Uncomment to use.
+            // const dateTimeValueInvalid = momentDateTimeValue.isBefore(moment('2015-01-01T10:00:00-00:00'))
+            // this.setProperties({
+            //     dateTimeValue,
+            //     dateTimeValueInvalid
+            // })
+            // if (!dateTimeValueInvalid) {
+            //     this.send('handleChange', dateTimeValue)
+            // }
+        },
+
     }
 
-    return {
-      date: '',
-      time: ''
-    }
-  },
-
-  // == Computed Properties ====================================================
-
-  // == Functions ==============================================================
-
-  parseValue (value) {
-    if (value.includes('-')) {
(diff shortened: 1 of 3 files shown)


addon/components/inputs/datetime.js
```diff
@@ -13,12 +14,40 @@ export default AbstractInput.extend({
...
+        date: arr[0] || '',
+        time: arr[1] || ''
+      }
+    }
+
+    return {
+      date: '',
+      time: ''
+    }
+  },
```
sandersky: Rather than maintaining state for the date and time I'd prefer to see you just use computed properties.

```js
/**
 * Get formatted datetime in a specific format
 * @param {String} value - datetime as a string
 * @param {String} format - format to represent datetime as
 * @returns {String} datetime in desired format
 */
function getFormattedDate (value, format) {
  if (typeOf(value) !== 'string') {
    return ''
  }

  const time = moment(value).format(format)
  return time === 'Invalid date' ? '' : time
}

export default AbstractInput.extend({

  …

  @readOnly
  @computed('value')
  date (value) {
    return getFormattedDate(value, 'YYYY:MM:DD')
  },

  @readOnly
  @computed('value')
  time (value) {
    return getFormattedDate(value, 'HH:mm:ss')
  }
```

Then in your template you can pass `date` and `time` directly into the two components.

addon/components/inputs/datetime.js
```diff
@@ -13,12 +14,40 @@ export default AbstractInput.extend({
...
   // == Functions ==============================================================
 
+  parseValue (value) {
+    if (value.includes('-')) {
+      this.set('date', value)
+    } else {
+      this.set('time', 'T' + value)
+    }
+    return this.get('date') + this.get('time')
+  },
```
sandersky: With the computed property change this can simply become:
```js
parseValue (value) {
  if (value.includes('-') {
    return `${this.get('date')}T${value}`
  }

  return `${value}T${this.get('time')}`
}
```

TheOtherDude:
Don't we need to test this?

sglanzer:
Which version of the date picker is this using?  This may be out of scope of the PR, but we've released a new version in line with the rest of the Frost LTS 1.0 lineup (7) and it may be worthwhile updating to that

addon/components/inputs/datetime.js
ZarifS: I've tried implementing your recommendation of computed properties but the functionality isn't the same. For example updating the time only is not possible, you would need to update the date and time. 
This is a problem for example if a user wants to search on time through Jan 1, 2016, they would need to put the date first then time, and then if they wish to update the time, they would need to reselect the date.. 
The code I placed before had both functionality, i can't seem to figure out how to make this work with computed..

addon/components/inputs/datetime.js
```diff
@@ -1,24 +1,55 @@
...
+    classNames: [
+        'frost-bunsen-input-datetime',
+        'frost-field'
+    ],
 
-  // == Computed Properties ====================================================
+    layout,
 
-  // == Functions ==============================================================
+    dateTimeValue: moment().subtract(1, 'day').subtract(1, 'hour').format(Format.dateTime),
```
sandersky: This doesn't work because in the case of an edit form it won't respect the value already in the form.

addon/components/inputs/datetime.js
```diff
@@ -1,24 +1,55 @@
...
+        onDateTimeChange(dateTimeValue) {
+
+            const momentDateTimeValue = moment(dateTimeValue)
+            if (!momentDateTimeValue.isValid()) {
+                this.set('dateTimeValueInvalid', true)
+                return 
+            }
+
+            //Comment this  else statment out if using with validation. 
+            else {
```
sandersky: When your if statement returns you don't need to wrap this logic in an else.

addon/components/inputs/datetime.js
```diff
@@ -1,24 +1,55 @@
...
+                this.send('handleChange', dateTimeValue)
+            }
+
+            // Example just to show validation - any date after 2015-01-01T10:00:00-00:00 is valid. Uncomment to use.
```
sandersky: Remove commented out code.

addon/components/inputs/datetime.js
sandersky: You really need to have a computed property to respect changes to the value outside of this renderer. Bunsen is designed to let consumers programmatically update the form value if they want and in order to do so when a consumer updates the value the renderer needs to take into account this new value.

ZarifS:
I still haven't completed the integration tests since after upgrading they need to be updated as well. Also fixed the code styling errors present in my last commit so you can disregard any you come across. Will include it in next commit. Since I included the upgrade for datepicker here I am also involved in making changes to the date renderer as well. Will include anything i make there onto my next commit as well.

sandersky:
@ZarifS I went ahead and got this working in this pull request https://github.com/ciena-frost/ember-frost-bunsen/pull/355/files

sglanzer:
@sandersky #355 doesn't include the upgrade though, correct?  I'd like to see 7.x used if possible since that version is the first to correctly use one way data flows (and is the LTS version)

sandersky:
@sglanzer I'd first like to get it working on 6.x since that is what bunsen uses. Then we can upgrade bunsen to 7.x and see if the tests all still pass and things still work, if not I'll need to make a new major of bunsen to upgrade to 7.x.

ZarifS:
@sandersky @sglanzer okay, i have the version with 7.0, so ill just stash my changes for those until its decided we want to upgrade to 7.0 renderer version

PR #356, opened 2017-02-09
Add template string support to endpoint select

job13er:

# CHANGELOG

* **Added** support for template strings in `endpoint` version of `select` renderer. You can now specify something like this in your bunsen view:  
  ```js 
  {
    endpoint: '/api/v1/heroes',
    recordsPath: 'data',
    labelAttribute: '${name} (${secret})'
    valueAttribute: 'id'
  }
  ```

* **Fixed** a bug where `record.get('title')` was being used instead of `get(record, 'title')` which broke in the ajax use case where the `record` was a POJO and did not have `.get()` defined.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,18 +32,19 @@ matrix:
 before_install:
   - npm config set spin false
   - npm install -g coveralls pr-bumper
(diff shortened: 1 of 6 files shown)


sandersky:
:+1:

sandersky:
:+1:

PR #357, opened 2017-02-09
Updated Datetime renderer with holding data and included integration tests. 

ZarifS:

# CHANGELOG

Added and updated integration tests to date time renderer. Took out some of the old tests that were not needed that were copied over from date renderer. Also includes change to fix bug that allows the renderer to properly hold data.

### DIFF
--- a/tests/dummy/app/pods/view/renderers/documentation/datetime.md
+++ b/tests/dummy/app/pods/view/renderers/documentation/datetime.md
@@ -3,7 +3,7 @@
 This renders a datetime input.
 
 ### Properties
-
+Currently disabled property is not supported.
 #### label
 
 ```json
--- a/tests/dummy/app/pods/view/renderers/documentation/datetime.md
+++ b/tests/dummy/app/pods/view/renderers/documentation/datetime.md
@@ -3,7 +3,8 @@
 This renders a datetime input.
 
 ### Properties
-Currently disabled property is not supported.
+Currently disabled property is not supported. Can disable inputs 
+by passing disabled = true to form but not through date-time-picker.
 #### label
 
 ```json
--- a/tests/dummy/app/pods/view/renderers/documentation/datetime.md
+++ b/tests/dummy/app/pods/view/renderers/documentation/datetime.md
@@ -1,0 +1,17 @@
+## datetime
+
+This renders a datetime input.
+
+### Properties
+
+#### label
+
+```json
+{
+  "label": "Bar",
+  "model": "foo",
+  "renderer": {
+    "name": "datetime"
+  }
+}
+```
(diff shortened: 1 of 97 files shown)


tests/integration/components/frost-bunsen-form/renderers/datetime-test.js
```diff
@@ -150,84 +154,6 @@ describe('Integration: Component / frost-bunsen-form / renderer / datetime', fun
     })
   })
 
-  describe('when form explicitly enabled', function () {
```
sandersky: Why did you delete all of these tests? It is important to make sure the renderer honors all of these various scenarios.

tests/integration/components/frost-bunsen-form/renderers/datetime-test.js
```diff
@@ -45,10 +46,15 @@ describe('Integration: Component / frost-bunsen-form / renderer / datetime', fun
   })
 
   it('renders as expected', function () {
-    expectCollapsibleHandles(0)
```
sandersky: Why was this removed?

tests/dummy/app/pods/view/renderers/documentation/datetime.md
```diff
@@ -3,7 +3,7 @@
 This renders a datetime input.
 
 ### Properties
-
+Currently disabled property is not supported.
```
sandersky: Disabled is a basic functionality bunsen expects. Consumers have cases where they disable forms while saving and in that scenario ALL inputs need to be disabled.

tests/integration/components/frost-bunsen-form/renderers/datetime-test.js
```diff
@@ -45,10 +46,15 @@ describe('Integration: Component / frost-bunsen-form / renderer / datetime', fun
...
     expectBunsenDatetimeRendererWithState('foo', {label: 'Foo'})
     expectOnValidationState(ctx, {count: 1})
   })
+  it('should have an input for date', function () {
+    expect($hook('bunsenForm-foo-datetimePicker-date-input').length).to.equal(1)
```
sandersky: I don't like seeing these expects in here. These should be covered under the test helpers for your custom renderer.

tests/integration/components/frost-bunsen-form/renderers/datetime-test.js
ZarifS: mainly since we didn't see why it was being checked each time but we can add it back no problem

tests/integration/components/frost-bunsen-form/renderers/datetime-test.js
ZarifS: Disabled property is not currently supported in frost-date-time picker, can be added back later when this becomes available, will add a sudo disabled to the template as discussed.

sandersky:
:+1:

PR #359, opened 2017-02-10
Fix CI build

job13er:

Since the last 2 PRs that were merged weren't published, #356 (a minor) and #357 (a patch), I'm including their changelogs here as well (although I reformatted the patch one to follow our guidelines more)

# CHANGELOG

* **Fixed** CI configuration to *hopefully* bump version at the right time now
* **Added** support for template strings in `endpoint` version of `select` renderer. You can now specify something like this in your bunsen view:  
  ```js 
  {
    endpoint: '/api/v1/heroes',
    recordsPath: 'data',
    labelAttribute: '${name} (${secret})'
    valueAttribute: 'id'
  }
  ```

* **Fixed** a bug where `record.get('title')` was being used instead of `get(record, 'title')` which broke in the ajax use case where the `record` was a POJO and did not have `.get()` defined. 

* **Added** and **Updated** integration tests to date time renderer. 
* **Removed** some of the old tests that were not needed that were copied over from date renderer. 
* **Fixed** a bug that prevented the renderer from properly holding data.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 dist: trusty
 language: node_js
 node_js:
-  - '6.9'
+  - '6.9.1'
   - 'stable'
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


job13er:
Approved

PR #366, opened 2017-02-23
Do not restrict IDs to numbers.

d-ashfield:
#PATCH#

# CHANGELOG

Remove restriction on IDs to be numbers when using queryForCurrentValue.

### DIFF
--- a/dependency-snapshot.json
+++ b/dependency-snapshot.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-frost-bunsen",
-  "version": "14.16.1",
+  "version": "14.16.2",
   "dependencies": {
     "5to6-codemod": {
       "version": "1.1.4",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-frost-bunsen",
-  "version": "14.16.1",
+  "version": "14.16.2",
   "description": "Create UI's from JSON configurations.",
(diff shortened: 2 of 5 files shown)


sandersky:
@d-ashfield rather than changing all of the existing tests that rely on a number you should add a new set of tests for the string case to ensure you didn't break the previous use case.

dependency-snapshot.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-frost-bunsen",
-  "version": "14.16.1",
+  "version": "14.16.2",
```
TheOtherDude: You don't need to update this. `pr-bumper` will take care of it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-frost-bunsen",
-  "version": "14.16.1",
+  "version": "14.16.2",
```
TheOtherDude: You don't need to update this. `pr-bumper` will take care of it.

sandersky:
@d-ashfield I noticed your pull request is from another branch on the upstream, `ciena-frost`, repository. Please don't do this and instead fork the repository, making any changes on your fork. Then submit pull requests from your fork against the upstream master branch. Otherwise Travis will see the commit as a PR being merged and publish a version bump commit, which you'll notice happened on this pull request.

PR #367, opened 2017-02-23
Adjust queryForCurrentValue to work with string IDs.

d-ashfield:
#PATCH#

# CHANGELOG

Modifying queryForCurrentValue to work with string IDs.


sandersky:
:+1:

PR #382, opened 2017-03-28
When renderer

juwara0:

Closes: https://github.com/ciena-frost/ember-frost-date-picker/issues/16

**Description:** A new "When" renderer that can be configured to allow for selecting a date via a keyword or a date time string:

```
Turn up
(x) Now
( ) specific date

Turn down
(x) Never
( ) specific date
```

**Requirements:** 
```
"When" renderer
"now" is selected by default
if "now" is selected disable date-time inputs
if "when" (specific date) is selected enable date-time inputs
date-time-picker defaults to current date/time

  configuration
    "now" label is configurable
    "now" value is configurable
    "when" label is not configurable
    "when" value is not configurable
    radio-button size is configurable and defaults to "small"
    dateFormat - is configurable and defaults to "YYYY-MM-DD"
    timeFormat - is configurable and defaults to "HH:mm:ss"
    dateTimeFormat - is configurable and defaults to "YYYY-MM-DDTHH:mm:ssZ"

When values are written to the Bunsen model
  The model property this field was mapped to will be set to one of two values:

    The "now" value supplied in the config as a "string"
    The "when" value of the timestamp "string" 
```
# CHANGELOG
* **Updated** date-time renderer test helpers to be more flexible so they can be used in other tests where date-time-picker is employed
* **Updated** date-time renderer tests to use the more flexible test helpers
* **Added** New `when` renderer that allows the selection of a date based on a keyword or a date time string.
* **Updated** demo app to showcase the new `when` renderer
* **Added** documentation for the new `when` renderer
* **Added** tests for the new `when` render

### DIFF
--- a/addon/components/inputs/when.js
+++ b/addon/components/inputs/when.js
@@ -54,6 +54,7 @@ export default AbstractInput.extend({
   /**
    * The date format to be used in the date-time-picker
    * @param {Object} cellConfig - contains the dateFormat "string" property (a valid moment format)
+   * @see {@link https://momentjs.com/docs/#/displaying/format}
    * @returns {String} the date format
    */
   dateFormat (cellConfig) {
@@ -65,6 +66,7 @@ export default AbstractInput.extend({
   /**
    * The time format to be used in the date-time-picker
    * @param {Object} cellConfig - contains the timeFormat "string" property (a valid moment format)
+   * @see {@link https://momentjs.com/docs/#/displaying/format}
    * @returns {String} the time format
    */
   timeFormat (cellConfig) {
@@ -76,6 +78,7 @@ export default AbstractInput.extend({
   /**
    * The date time format to be used for formating the value sent to the bunsen model
    * @param {Object} cellConfig - contains the dateTimeFormat "string" property (a valid moment format)
+   * @see {@link https://momentjs.com/docs/#/displaying/format}
    * @returns {String} the date time format
    */
   dateTimeFormat (cellConfig) {
--- a/addon/components/inputs/when.js
+++ b/addon/components/inputs/when.js
@@ -1,0 +1,198 @@
+import Ember from 'ember'
+const {$, get, guidFor, run} = Ember
+import computed, {readOnly} from 'ember-computed-decorators'
+import {PropTypes} from 'ember-prop-types'
+import moment from 'moment'
+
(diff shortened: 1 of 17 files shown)


tests/dummy/app/pods/view/renderers/documentation/when.md
```diff
@@ -0,0 +1,101 @@
...
+    dateTimeFormat: 'MM-DD-YYYYTHH:mm:ssZ',
+    name: 'when'
+  }
+}
+```
+
+#### renderer.label
+
+The label for the first radio button
+```json
```
agonza40: Nit-picky, but these blocks aren't valid JSON.

addon/components/inputs/when.js
```diff
@@ -0,0 +1,201 @@
...
+
+    this.setProperties({
+      date,
+      firstButtonValue,
+      selectedValue: firstButtonValue,
+      storedDateTimeValue: moment(currentDateTime).format(this.get('dateTimeFormat')),
+      time
+    })
+
+    run.later(() => {
```
agonza40: Why is this necessary? When did the value change?

tests/dummy/app/pods/view/renderers/documentation/when.md
juwara0: I have made these updates.

addon/components/inputs/when.js
cstolli: I'm not sure it is necessary either, pretty sure doing a `setProperties` in `init` does not trigger a re-render anyway, but there might be some reason I'm not thinking of to exhaust the runLoop before calling the `onChange` here, hence the `run.later`

cstolli:
approved

cstolli:
👍

addon/components/inputs/when.js
juwara0: This is necessary since the default is to have the first radio button selected we also need to write that value to the bunsen model.

Without wrapping the setting of the initial value to the bunsen model, I get an Ember error that `renderValue` is being modified twice. Once in `component:frost-bunsen-cell` and another in this renderer: `component:frost-bunsen-input-when`.

cstolli: yeah makes sense.

PR #389, opened 2017-04-19
Added check blueprint

dafortin:

# CHANGELOG
* **Added** check in the blueprint to avoid installing packages already installed


package.json
```diff
@@ -104,7 +104,7 @@
     "ember-bunsen-core": "0.24.0",
     "ember-cli-babel": "^5.1.8",
     "ember-cli-htmlbars": "^1.0.3",
-    "ember-frost-core": "1.12.0",
+    "ember-frost-core": "1.13.0",
```
sandersky: Changing the min version of a dependency should not happen in a patch. Since it isn't a major version bump of `ember-frost-core` I'd say it is safe to make this PR a minor.

job13er: Seems wrong to have a pinned version in `dependencies` too. Is it true that this addon will break if the consuming app is using a later version? If not, it should probably allow a more recent, version-compatible code to be installed. Otherwise you'll end up with two copies in `node_modules` and I'm not sure which one lands in the app.

sandersky: It actually has a caret now, I fixed it earlier today. Just want to make sure the minimum version isn't changed here.

dafortin: Fixed

blueprints/ember-frost-bunsen/index.js
```diff
@@ -1,10 +1,10 @@
...
-    {name: 'ember-frost-date-picker', target: '^6.0.0'},
-    {name: 'ember-frost-fields', target: '^4.0.0'},
-    {name: 'ember-frost-table', target: '^1.0.0'}
-  ]
-}
+const blueprintHelper = require('ember-frost-core/blueprint-helper')
+
+const addonsToAdd = [
+  {name: 'ember-frost-date-picker', target: '^6.0.0'},
+  {name: 'ember-frost-fields', target: '^4.0.0'},
```
sglanzer: Bunsen doesn't use `core` directly?  It's all through `fields`?

dafortin: No it's using core. I noticed `frost-button`

job13er: I *think* it only includes fields because it uses the CSS

sandersky:
:+1:

PR #391, opened 2017-04-20
Select basic types support

### LINKED ISSUE #263: Support non-strings in the select renderer
It would be super awesome if the select renderer didn't assume the value was always a string. If the values are numbers, there are some cases where setting a default that is `0` doesn't work. Also having 0 doesn't show up in the drop down list.  When the `enum` is overridden in renderer options, it gets further by displaying the `0` but it still doesn't allow it as a default.

model:
```json
{
  "type": "object",
  "properties": {
    "pcp": {
      "type": "number",
      "enum": [0, 1]
    }
  }
}
```

view: 
```json
{
  "cells": [
    {
      "model": "pcp"
    }
  ],
  "type": "form",
  "version": "2.0"
}
```

### LINKED ISSUE #387: Update select endpoint to support pointing to primitive arrays
Currently the select renderers endpoint property expects an array of objects for populating the select. We should extending this to also allow pointing to an array of strings or numbers.

sandersky:

resolves #263 
resolves #387 

# CHANGELOG

* **Added** support for select renderer to be driven by an array of integers/numbers/strings when using `endpoint` or simply `recordsPath` to mine the form value.


sandersky:
:+1:

PR #393, opened 2017-04-21
Fix issue where array label doesn't show

### LINKED ISSUE #392: Missing array label when arrayOptions not present
[This](https://github.com/ciena-frost/ember-frost-bunsen/blob/4f4af6b11cc671f63df6d30c14f7ad1042dae014/addon/components/cell.js#L324) is making the assumption that `arrayOptions` is always defined for a cell representing an array. In actuality you can provide an array without defining `arrayOptions` and bunsen will still render it. However in this scenario bunsen will fail to include the array label/header for you. We should improve this CP to actually check type type of the model property and make more informed decisions based on that.

*Model*

```json
{
  "properties": {
    "foo": {
      "items": {
        "type": "string"
      },
      "type": "array"
    }
  },
  "type": "object"
}
```

*View (Missing Label)*

```json
{
  "cells": [
    {
      "label": "Test",
      "model": "foo"
    }
  ],
  "type": "form",
  "version": "2.0"
}
```

*View (Shows Label)*

```json
{
  "cells": [
    {
      "label": "Test",
      "model": "foo",
      "arrayOptions": {}
    }
  ],
  "type": "form",
  "version": "2.0"
}
```

sandersky:

resolves #392 

# CHANGELOG

* **Fixed** bug where array label wasn't showing up when `arrayOptions` is not defined.

### DIFF
--- a/addon/components/cell.js
+++ b/addon/components/cell.js
@@ -316,12 +316,12 @@ export default Component.extend(HookMixin, PropTypeMixin, {
   },
 
   @readOnly
-  @computed('cellConfig')
-  showSection (cellConfig) {
+  @computed('cellConfig', 'subModel.type')
+  showSection (cellConfig, type) {
     return (
       cellConfig.collapsible ||
       (cellConfig.label && cellConfig.children) ||
-      (cellConfig.arrayOptions && !cellConfig.hideLabel)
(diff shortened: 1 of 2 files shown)


sandersky:
:+1:

PR #399, opened 2017-05-02
add property for disabling validation on visibility change

PoplarToppler:

The motivation behind this is that when using the form for facets with validation, I am using onValidation to trigger the server request so that I am able to block invalid filters (e.g. invalid date range) while still allowing the query to succeed. 

The problem is that if we want to freeze the current view (e.g. stop polling), then go to another tab (e.g. new details tab) then return to the original view, a new query is triggered due to the validation of the form being triggered on visibility change. This ends up changing the view that was supposed to be frozen.

# CHANGELOG
* **Added** validateOnVisibilityChange property to allow disabling of form validation after losing/regaining focus of the page
* **Fixed** select documentation formatting

### DIFF
--- a/tests/unit/components/frost-bunsen-form/component-test.js
+++ b/tests/unit/components/frost-bunsen-form/component-test.js
@@ -18,7 +18,6 @@ describe('Unit: frost-bunsen-form', function () {
       onChange () {},
       onError () {}
     })
-    component.init()
   })
 
   afterEach(function () {
--- a/tests/unit/components/frost-bunsen-form/component-test.js
+++ b/tests/unit/components/frost-bunsen-form/component-test.js
@@ -1,0 +1,109 @@
+import {expect} from 'chai'
+import {setupComponentTest} from 'ember-mocha'
+import {afterEach, beforeEach, describe, it} from 'mocha'
+import sinon from 'sinon'
+
+describe('Unit: frost-bunsen-form', function () {
+  setupComponentTest('frost-bunsen-form', {
+    unit: true
+  })
+
+  let component, sandbox
+
+  beforeEach(function () {
+    sandbox = sinon.sandbox.create()
+    component = this.subject({
+      bunsenModel: {},
+      bunsenView: {},
+      onChange () {},
+      onError () {}
+    })
+    component.init()
+  })
+
+  afterEach(function () {
+    sandbox.restore()
+  })
+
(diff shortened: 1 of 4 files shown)


PoplarToppler:
Let me know if you would like me to add any tests around this. I didn't see any existing ones so I wasn't sure where to put them if I should add any.

sandersky:
@PoplarToppler it'd be nice to see some tests, even if it is just a few unit tests against the `_onVisiblityChange` method for the various cases (probably 3 tests in total).

PoplarToppler:
Sorry I have been super busy. I will get back to and finish this soon.

PoplarToppler:
I'm not sure where I should put these tests, @job13er can you give me a recommendation? Thanks!

job13er:
I'm not really familiar with the testing here. I'd defer to @agonza40 or @sophypal.

job13er:
I've also pretty much stopped using bunsen now as well (no more data driven requirements) so I'm not sure I'm the best person to review this. Do we have someone who is going to step up as primary maintainer with Matt gone?

sglanzer:
We will, but we're not in a position to do it until later in June

PoplarToppler:
@agonza40 @sophypal do either of you have suggestions of where I could put a few tests for this new property? I'd ideally like to get this in soon. My best guess is I'd add a new file: tests/unit/components/frost-bunsen-form/on-visibility-change-test.js and go from there. Does that sound acceptable?

sophypal:
@PoplarToppler Typically we model the test directory against the src directory. Since there is very little unit test (due to favoring integration in this project), you'd need to add `tests/unit/components/frost-bunsen-form/component-test.js` and have your scenario in it's own `describe` block.

tests/unit/components/frost-bunsen-form/component-test.js
```diff
@@ -0,0 +1,109 @@
...
+
+  beforeEach(function () {
+    sandbox = sinon.sandbox.create()
+    component = this.subject({
+      bunsenModel: {},
+      bunsenView: {},
+      onChange () {},
+      onError () {}
+    })
+    component.init()
```
sophypal: You shouldn't need to call `init` here as `this.subject` calls it indirectly through instantiation.

sophypal:
👍

PR #400, opened 2017-05-04
Add more waits and remove some deep cloning

sandersky:

# CHANGELOG

* **Added** missing `return wait()` call to more tests to make sure they are async safe.
* **Replaced** some deep cloning with shallow cloning to reduce new object creation.

### DIFF
--- a/addon/utils.js
+++ b/addon/utils.js
@@ -1,7 +1,6 @@
 import Ember from 'ember'
 const {get, merge, typeOf} = Ember
 import config from 'ember-get-config'
-import _ from 'lodash'
 
 const {keys} = Object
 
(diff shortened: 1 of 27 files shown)


addon/utils.js
```diff
@@ -160,20 +159,27 @@ export function getMergedConfig (cellConfig, cellDefinitions) {
...
-    const mergedChildConfigs = []
-    mergedConfig.children.forEach((childConfig) => {
-      mergedChildConfigs.push(getMergedConfigRecursive(childConfig, cellDefinitions))
-    })
-    mergedConfig.children = mergedChildConfigs
+    const children = mergedConfig.children.map((childConfig) => getMergedConfigRecursive(childConfig, cellDefinitions))
+    const childrenChanged = children.some((child, index) => child !== mergedConfig.children[index])
+
+    if (childrenChanged) {
+      mergedConfig = Object.assign({}, mergedConfig, {children})
```
agonza40: Is it necessary to create a new copy of this object?

sandersky: This is keeping the function pure. Otherwise it can end up mutating the functions input(s). `mergedConfig` isn't actually a new object when the `cellConfig` passed into `getMergedConfig` doesn't contain the `extends` property.

sandersky:
:+1:

PR #407, opened 2017-05-10
Add internal state support

sandersky:

# CHANGELOG

* **Added** support for new `_internal` model property which can contain any properties you don't want to propagate back to the consumer via the `onChange` callback. These can be used to drive conditional views without polluting the form value with state.
* **Fixed** `geolocation` renderer to make less set calls to tighten the Ember run loop around changes.
* **Replaced** some `didReceiveAttrs()` hooks with computed properties since Ember has deprecated the use of the `attrs` argument in the `didReceiveAttrs()` life cycle hook.


sandersky:
:+1:

PR #408, opened 2017-05-10
Validate when updating

sandersky:

# CHANGELOG

* **Cleaned** up code to run validation in same run loop as setting properties for optimize components a bit more.
* **Fixed** section component to properly apply class for expand/collapse icon.
* **Fixed** tests to clear out variables defined in `describe` blocks to prevent memory leaks.


cstolli:
👍

PR #417, opened 2017-05-18
Fix sever requests when filtering multi-select dropdown

CanJee:

When filtering in the multi-select dropdown, it would not send requests to the server with the filter query but just filter locally. In select dropdown, this functionality works correctly. Requests are sent to the server every time you type something in the filter field.

# CHANGELOG
- Fix server requests not being sent when filtering on the multi-select dropdown

### DIFF
--- a/addon/templates/components/frost-bunsen-input-multi-select.hbs
+++ b/addon/templates/components/frost-bunsen-input-multi-select.hbs
@@ -13,8 +13,8 @@
     disabled=disabled
     error=(if renderErrorMessage true false)
     hook=hook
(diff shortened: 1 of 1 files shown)


job13er:
Can you please add a test case that exercises this, so that we make sure it doesn't regress?

Thanks.

TheOtherDude:
I've got this covered with a test in PR#419.

I'll even add another test to make sure the filter action is called :)

CanJee:
@TheOtherDude great thanks! I will close this PR then.

PR #418, opened 2017-05-22
Fixed the way view cells are merged.

agonza40:

# CHANGELOG
**Fixed** cell definition re-use. In certain situations cell definitions were being used directly instead of being copied which caused issues when trying to assign an ID to the result cell. Now cell definitions are copied instead of being used directly.

### DIFF
--- a/addon/utils.js
+++ b/addon/utils.js
@@ -165,8 +165,14 @@ export function getMergedConfigRecursive (cellConfig, cellDefinitions) {
 
   // recursive object case
   if (mergedConfig.children) {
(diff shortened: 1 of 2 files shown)


cstolli:
👍

addon/utils.js
```diff
@@ -184,7 +190,7 @@ export function getMergedConfigRecursive (cellConfig, cellDefinitions) {
           if (itemCell !== mergedConfig) {
             itemCellChanged = true
           }
-          return mergedConfig
+          return Object.mergedConfig
```
sandersky: This doesn't make sense to me.

PR #419, opened 2017-05-23
Pass selectSpreadOptions to multi-select renderer

TheOtherDude:
### This project uses [semver](semver.org), please check the scope of this pr:

- [ ] #none# - documentation fixes and/or test additions
- [x] #patch# - backwards-compatible bug fix
- [ ] #minor# - adding functionality in a backwards-compatible manner
- [ ] #major# - incompatible API change

The main purpose of this PR is to allow API filtering in multi-selects. Currently, it always uses local filtering because `onInput` is undefined.

# CHANGELOG

* **Fixed** multi-select renderer not receiving renderer options from view
* **Fixed** multi-select renderer API filtering

PR #422, opened 2017-05-25
Fix multi-select renderer

job13er:

Another attempt to get #419 through. 

# CHANGELOG
* **Fixed** specifying `options` in the bunsen view for a `multi-select` renderer. The `options` were not making it all the way to the `frost-multi-select` component. 
* **Fixed** API filtering for multi-select. Previously, it was only using local filtering.

### DIFF
--- a/tests/integration/components/frost-bunsen-form/renderers/multi-select/options/ember-data-view-query-test.js
+++ b/tests/integration/components/frost-bunsen-form/renderers/multi-select/options/ember-data-view-query-test.js
@@ -11,39 +11,43 @@ import sinon from 'sinon'
 import {
   expectBunsenInputToHaveError,
   expectCollapsibleHandles,
-  expectOnChangeState,
   expectOnValidationState
 } from 'dummy/tests/helpers/ember-frost-bunsen'
 
 import {expectSelectWithState} from 'dummy/tests/helpers/ember-frost-core'
 import selectors from 'dummy/tests/helpers/selectors'
 
-describe('Integration: Component / frost-bunsen-form / renderer / select Ember Data view query /', function () {
+describe.only('Integration: Component / frost-bunsen-form / renderer / multi-select Ember Data view query /', function () {
   setupComponentTest('frost-bunsen-form', {
     integration: true
   })
 
-  let props, sandbox, resolver
+  let props, sandbox, resolver, store
 
   beforeEach(function () {
     sandbox = sinon.sandbox.create()
     initialize()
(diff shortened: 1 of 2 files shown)


tests/integration/components/frost-bunsen-form/renderers/select/options/ember-data-view-query-test.js
```diff
@@ -452,13 +452,13 @@ describe('Integration: Component / frost-bunsen-form / renderer / select Ember D
                   options: {
                     labelAttribute: 'label',
                     modelType: 'node',
+                    placeholder: 'Foo bar',
```
job13er: This wasn't strictly necessary, I just wanted to make sure it worked here before trying it in the `multi-select`

tests/integration/components/frost-bunsen-form/renderers/multi-select/options/ember-data-view-query-test.js
```diff
@@ -0,0 +1,1291 @@
+import {expect} from 'chai'
```
job13er: This file is so huge because I started out with a `select` renderer test and then just switched it to `multi-select` and removed bits that weren't appropriate for a multi-select. It's probably more tests than we need for this fix, but they should probably be here anyway.

tests/integration/components/frost-bunsen-form/renderers/multi-select/options/ember-data-view-query-test.js
```diff
@@ -0,0 +1,1291 @@
...
+          this.set('bunsenView', {
+            cells: [
+              {
+                model: 'foo',
+                renderer: {
+                  name: 'multi-select',
+                  options: {
+                    labelAttribute: 'label',
+                    modelType: 'node',
+                    placeholder: 'Foo bar',
```
job13er: Here's the test that actually exercises the bug.

tests/integration/components/frost-bunsen-form/renderers/multi-select/options/ember-data-view-query-test.js
```diff
@@ -0,0 +1,1291 @@
...
+              text: ''
+            })
+
+            expectOnValidationState({props}, {count: 0})
+          })
+        })
+      })
+    })
+
+    describe('when user types', function () {
```
job13er: And here's the test that covers #417

TheOtherDude:
:+1:

PR #423, opened 2017-05-30
Select renderer relative path support for recordsPath option

agonza40:

# CHANGELOG
**Added** support for relative paths for the recordsPath select renderer option
**Fixed** bunsenId, bunsenModel, and config assignment to cells from the inline array item component.


cstolli:
I think we need some tests on this stuff?

cstolli:
👍

PR #426, opened 2017-06-02
Multi-select only filters locally when item selected

CanJee:

Partially fixed in https://github.com/ciena-frost/ember-frost-bunsen/pull/422 but we did not consider this case.

# CHANGELOG
- **Fixed** issue where when you select an item in the `multi-select` dropdown then filter it only filters locally.


TheOtherDude:
:+1:

PR #429, opened 2017-06-05
Multi-select selected items disappear when filtering fix

CanJee:

Now that we have API filtering working as expected for select/multi-select dropdowns, we need to address an issue with the multi-select dropdown. Currently, when you are filtering a multi-select dropdown, the selected items disappear if they do not match the filter term. This is not what was intended. Instead we would like the selected items to always remain at the top when filtering. That is what has been implemented in this PR. This change can be seen locally by navigating to:

`http://localhost:4200/#/examples?model=select&view=select-form`

# CHANGELOG
- **Fixed** issue where filtering in multi-select caused selected items to disappear
- **Added** a mirage endpoint for failing API requests
- **Added** tests for this fix


CanJee:
Closed for now requirements are still not clear.

PR #432, opened 2017-06-14
Internal Models and model expansions.

agonza40:

# CHANGELOG
**Changed** model expansions and internal models to take effect in nested objects.
**Added** support for internal models within arrays.

### DIFF
--- a/addon/components/detail.js
+++ b/addon/components/detail.js
@@ -246,7 +246,9 @@ export default Component.extend(SpreadMixin, HookMixin, PropTypeMixin, {
   applyStoreUpdate ({lastAction, newProps, validationResult, value}) {
     if (Object.keys(newProps).length !== 0) {
       const model = newProps.renderModel || this.get('renderModel')
-      const view = newProps.view ? this.getRenderView(model, newProps.view) : this.getRenderView(model, this.get('view'))
+      const view = newProps.view
+        ? this.getRenderView(model, newProps.view)
+        : this.getRenderView(model, this.get('view'))
 
       Object.assign(newProps, this.validateSchemas(model, view))
 
--- a/addon/components/detail.js
+++ b/addon/components/detail.js
@@ -9,6 +9,8 @@ import {
   viewV1ToV2
 } from 'bunsen-core'
 
+import {clearInternals} from 'bunsen-core/utils'
+
 const {
   CHANGE_VALUE,
   changeModel,
@@ -31,6 +33,7 @@ const createStoreWithMiddleware = applyMiddleware(thunkMiddleware)(createStore)
 
 import {
   deemberify,
+  findInternalValues,
   getMergedConfigRecursive,
   isRegisteredEmberDataModel,
   validateRenderer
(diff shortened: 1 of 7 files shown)


addon/components/detail.js
```diff
@@ -242,7 +246,9 @@ export default Component.extend(SpreadMixin, HookMixin, PropTypeMixin, {
   applyStoreUpdate ({lastAction, newProps, validationResult, value}) {
     if (Object.keys(newProps).length !== 0) {
       const model = newProps.renderModel || this.get('renderModel')
-      const view = newProps.view ? this.getRenderView(model, newProps.view) : this.get('renderView')
+      const view = newProps.view
+        ? this.getRenderView(model, newProps.view)
+        : this.getRenderView(model, this.get('view'))
```
sophypal: Does this mean that the view normalization is performed all the time on every store update?

addon/utils.js
```diff
@@ -404,3 +405,32 @@ export function isModelPathValid (path, bunsenModel) {
...
+ * @param {any} val Value to check for internal values
+ * @returns {String[]} List of strings that are paths to internal values
+ */
+export function findInternalValues (val) {
+  return _.chain(val)
+    .map(function (item, key) {
+      if (key === '_internal') {
+        return key
+      }
+      if (['boolean', 'number', 'string', 'undefined', 'null'].indexOf(typeof item) !== -1) {
```
sophypal: I've been using `[].includes` to check existence in an array more than `indexOf`.

addon/utils.js
```diff
@@ -404,3 +405,32 @@ export function isModelPathValid (path, bunsenModel) {
...
+        return
+      }
+      const subPaths = findInternalValues(item)
+      if (subPaths.length <= 0) {
+        return
+      }
+      return _.map(subPaths, function (subPath) {
```
sophypal: Any particular reason you're favoring `_.map` over `map`?

addon/utils.js
```diff
@@ -404,3 +405,32 @@ export function isModelPathValid (path, bunsenModel) {
...
+        return `${key}.${subPath}`
+      })
+    })
+    .filter()
```
sophypal: So what does a `filter` with no args do?

addon/components/detail.js
agonza40: Yeah, I was having trouble getting the new normalized view from the computed property in the cases where I needed it. I think this can all be smoothed out with some slight restructuring of this component.

addon/utils.js
agonza40: That's what I was looking for! I was trying to use `[].contains()` which isn't a real thing.

addon/components/detail.js
sophypal: Ok, lets hope this doesn't cause a performance regression. Can you double-check that the EVC example still runs smoothly with this change?

sophypal:
👍

addon/utils.js
agonza40: Filters out falsey values.

addon/utils.js
agonza40: I may have originally thought `subPaths` could be something other than an array.

sophypal:
👍

PR #433, opened 2017-06-15
Fix queryForCurrentValue

sophypal:

# CHANGELOG

* **Fixed** a bug with the `queryForCurrentValue` select option which assumed the id was at the top level of the form's value.

### DIFF
--- a/addon/list-utils.js
+++ b/addon/list-utils.js
@@ -135,7 +135,7 @@ export function getItemsFromAjaxCall ({ajax, bunsenId, data, filter, options, va
 export function getItemsFromEmberData (value, modelDef, data, bunsenId, store, filter) {
   const modelType = modelDef.modelType || 'resources'
   const {labelAttribute, queryForCurrentValue, valueAttribute} = modelDef
-  const valueAsId = value[bunsenId]
+  const valueAsId = get(value, bunsenId)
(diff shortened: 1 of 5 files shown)


tests/unit/list-utils-test.js
```diff
@@ -663,7 +667,7 @@ describe('Unit: list-utils', function () {
           valueAttribute: 'secret',
           query: {
             booleanFlag: true,
-            universe: '${../universe}'
+            universe: '${./universe}'
```
sophypal: `${./universe}` is correct as the `findValue` utility in `bunsen-core` doesn't behave like the path format followed by `conditions`. Oddly enough, both `../` and `./` will both succeed when the value object is one-level deep.

addon/list-utils.js
```diff
@@ -135,7 +135,7 @@ export function getItemsFromAjaxCall ({ajax, bunsenId, data, filter, options, va
 export function getItemsFromEmberData (value, modelDef, data, bunsenId, store, filter) {
   const modelType = modelDef.modelType || 'resources'
   const {labelAttribute, queryForCurrentValue, valueAttribute} = modelDef
-  const valueAsId = value[bunsenId]
```
sophypal: I have to believe this was an oversight as most models aren't this flat.

sophypal:
👍

PR #438, opened 2017-06-19
Cherry picking commits from 14.23.x branch

sophypal:

# CHANGELOG

* **Fixed** a bug with the queryForCurrentValue select option which assumed the id was at the top level of the form's value.
* **Fixed** `components/form.js` to allow extension of `classNames` rather than hard-setting it to only one class name.


agonza40:
👍

PR #445, opened 2017-06-21
Internal model/property fixes

agonza40:

# CHANGELOG
**Fixed** removal of `_internal` properties for `onChange` handler.
**Updated** bunsen-core to fix validation for deep internal properties

### DIFF
--- a/addon/utils.js
+++ b/addon/utils.js
@@ -2,6 +2,7 @@ import Ember from 'ember'
 const {get, merge, typeOf} = Ember
 import config from 'ember-get-config'
 import _ from 'lodash'
+import immutable from 'seamless-immutable'
 
 const {keys} = Object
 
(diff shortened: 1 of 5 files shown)


addon/utils.js
```diff
@@ -407,30 +408,55 @@ export function isModelPathValid (path, bunsenModel) {
...
+  if (!Array.isArray(val)) {
+    let withoutInternal
+    if (immutable.isImmutable(val)) {
+      withoutInternal = val.without('_internal')
+      const childrenWithoutInternals = removeChildInternals(withoutInternal)
+      return withoutInternal.merge(childrenWithoutInternals)
+    }
+
+    if (val._internal !== undefined) {
+      withoutInternal = Object.assign({}, val)
```
sophypal: Would be cool to use that same logic for `Immutables` using the `without` for non-Immutable objects as well. Can you achieve the same with Lodash's version. I just think it would be easier to follow.

sophypal:
👍

addon/utils.js
agonza40: Actually, I think I can do that pretty easy

addon/utils.js
```diff
@@ -440,20 +440,26 @@ export function removeInternalValues (val) {
...
-      withoutInternal = Object.assign({}, val)
-      delete withoutInternal._internal
+      merge = immutable.merge
+      without = immutable.without
+    } else if (val._internal !== undefined) {
+      merge = Object.assign
+      without = _.omit
     } else {
-      withoutInternal = val
+      merge = function (first, second) {
```
sophypal: Would this local `merge` implementation be better suited in external to this function. Nested functions do have some performance drawbacks especially inside something that will be recursed. A new lexical environment (closure) needs to be created for every recursive step.

addon/utils.js
```diff
@@ -440,20 +440,26 @@ export function removeInternalValues (val) {
...
+        if (Object.keys(second).length > 0) {
+          Object.assign({}, first, second) // don't mutate the object
```
sophypal: Um, does this even work since you're not returning from the `assign`.

sophypal: Also, what if the first was bigger then the second? Wouldn't you want to merge it still?

agonza40: Thanks

agonza40:
👍

PR #447, opened 2017-06-22
Update ember-frost-date-picker to `^7.1.0`

dafortin:

screen shot 2017-06-22 at 3 26 15 pm
screen shot 2017-06-22 at 3 26 23 pm
screen shot 2017-06-22 at 3 26 35 pm

# CHANGELOG
* **Updated** `ember-frost-date-picker@^7.1.0`

### DIFF
--- a/addon/components/inputs/datetime.js
+++ b/addon/components/inputs/datetime.js
@@ -4,6 +4,10 @@ import moment from 'moment'
 import AbstractInput from './abstract-input'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-datetime'
 
+const DATE_FORMAT = 'YYYY-MM-DD'
+const TIME_FORMAT = 'HH:mm:ss'
+const DATE_TIME_FORMAT = 'YYYY-MM-DDTHH:mm:ssZ'
+
 export default AbstractInput.extend({
   // == Component Properties ===================================================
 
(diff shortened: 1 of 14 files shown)


addon/components/inputs/when.js
```diff
@@ -167,22 +166,24 @@ export default AbstractInput.extend({
...
-      // Needed because clicking in the date-time-picker changes the event object.
-      if (event.target.value === DATE_VALUE || event.target.value === firstButtonValue) {
-        this.set('selectedValue', event.target.value)
-      }
-
       // Set the bunsen model to the value of the selected button
       this.onChange(
         this.get('bunsenId'),
-        (event.target.value === firstButtonValue) ? firstButtonValue : this.get('storedDateTimeValue')
+        (value === firstButtonValue) ? firstButtonValue : (
```
sophypal: This is harder to read then the previous implementation.

tests/integration/components/frost-bunsen-form/renderers/when-test.js
```diff
@@ -83,7 +83,7 @@ describe('Integration: Component / frost-bunsen-form / renderer / when', functio
     it('renders as expected', function () {
       expectCollapsibleHandles(0)
       expectBunsenWhenRendererWithState('foo', {label: 'FooBar Baz'})
-      expectOnValidationState(ctx, {count: 2})
+      expectOnValidationState(ctx, {count: 4})
```
sophypal: I think this means that 2 more validations resulted from this change? How? I didn't see any change that would have contributed to this.

addon/components/inputs/when.js
dafortin: It's not exactly the same as previous implementation. I can try to make it clearer.

tests/integration/components/frost-bunsen-form/renderers/when-test.js
dafortin: The implementation of the when function is a bit different and that's why we are getting more validation.

sophypal: Makes sense.

sophypal:
👍

addon/components/inputs/datetime.js
```diff
@@ -21,21 +25,30 @@ export default AbstractInput.extend({
...
+      return moment().format(DATE_TIME_FORMAT)
+    }
+    return value
+  },
+
   // == Functions ==============================================================
 
   parseValue (value) {
-    return value.format('YYYY-MM-DDTHH:mm:ssZ')
+    return moment(value).format(DATE_FORMAT)
```
sophypal: Good catch!

PR #454, opened 2017-07-04
Don't display section label for renderers

sophypal:

# CHANGELOG

* **Fixes** a regression in labeling that would always display labels if the model was an array. It's been modified to not display the label if the cell has a renderer for the it.


sophypal:
👍

Issue #365, opened 2017-02-23
Allow sorting of query results in the select renderer 


sglanzer:
Via function and/or keyword

sandersky:
I'm not sure bunsen should be concerned with this. I think this should be fixed at the API level or in the Ember Data adapter.

PR #455, opened 2017-07-04
Allow sorting of query results in the select renderer #365

durbose:

# CHANGELOG
- Allow sorting of query results in the select renderer


sophypal:
@durbose Thanks for the PR. I'd prefer if the sort was available on the select component, not the renderer. The renderer can pass whatever options you want but this feature would be beneficial to the select component itself inside of frost-core.

PR #460, opened 2017-07-26
Update to ember-cli@2.12.3

quincyle:

# CHANGELOG
* **Update** to `ember-cli@2.12.3`


.travis.yml
```diff
@@ -15,11 +15,11 @@ addons:
   firefox: 'latest-esr'
 cache:
   directories:
-  - node_modules
+  - node_module
```
juwara0: Should be `node_modules`

EWhite613:
Please refer to https://github.com/ciena-frost/ember-frost-core/pull/480#issuecomment-318709015

EWhite613:
approved

EWhite613:
Approved

juwara0:
DO NOT MERGE YET.

tests/unit/components/inputs/select-test.js
```diff
@@ -9,7 +9,9 @@ import utils from 'ember-frost-bunsen/list-utils'
 
 const {run} = Ember
 
-describe('Unit: frost-bunsen-input-select', function () {
+// Quincy Le 2017-07-30
+// can't find injection service:ajax with ember-cli: 2.12.3
+describe.skip('Unit: frost-bunsen-input-select', function () {
   setupComponentTest('frost-bunsen-input-select', {
```
EWhite613: Change this to 
```
setupComponentTest('frost-bunsen-input-select', {
    unit: true,
    needs: ['service:ajax']
  })
```
Then the service will be injected

quincyle: Already tried it with no luck.

package.json
```diff
@@ -36,55 +36,51 @@
...
     "ember-disable-prototype-extensions": "^1.1.0",
-    "ember-disable-proxy-controllers": "^1.0.1",
-    "ember-elsewhere": "0.4.1",
-    "ember-export-application-global": "^1.1.1",
-    "ember-frost-core": "1.14.3",
+    "ember-elsewhere": "~1.0.1",
+    "ember-export-application-global": "^1.0.5",
     "ember-frost-date-picker": "^7.1.0",
-    "ember-frost-demo-components": "^3.0.1",
+    "ember-frost-demo-components": "git://github.com/helloWorldCindy/ember-frost-demo-components.git#upgrade",
```
juwara0: This can be updated to now use the published version `3.1.2`

config/ember-try.js
```diff
@@ -1,140 +1,36 @@
...
-        }
-      }
-    },
-    {
-      name: 'ember-2-3',
+      name: 'ember-2-11',
       bower: {
         dependencies: {
-          'ember': '~2.3.0'
+          'ember': '~2.11.0'
```
juwara0: This needs to be the setup for version 2.8

.travis.yml
```diff
@@ -13,13 +13,10 @@ addons:
...
       - g++-4.8
   firefox: 'latest-esr'
-cache:
-  directories:
-  - node_modules
 env:
   matrix:
-    - EMBER_TRY_SCENARIO=ember-2-8
-    - EMBER_TRY_SCENARIO=default
+    - EMBER_TRY_SCENARIO=ember-2-11
```
juwara0: This needs to be running ember-2-8 instead.

quincyle: As we've discussed, in the previous version, Bunsen actually never gets run with `ember: 2.8` (With `ember-source` not being set as null in config/ember-try, CI will always run against ember-source presented in package.json). I tried to run it against `ember: 2.8.0`, but I'm seeing extra test failure. Since it never runs against 2.8, I won't bother to fix these tests at this moment.

tests/unit/components/inputs/select-test.js
quincyle: Turns out ember treats `'service:ajax'` and `'service: ajax'` differently. `'service:ajax'` is the correct syntax.

package.json
```diff
@@ -36,55 +36,51 @@
...
     "ember-browserify": "1.1.13",
-    "ember-cli": "~2.11.0",
-    "ember-cli-app-version": "^2.0.1",
+    "ember-cli": "2.12.3",
     "ember-cli-chai": "^0.3.2",
     "ember-cli-code-coverage": "0.3.11",
     "ember-cli-dependency-checker": "^1.3.0",
-    "ember-cli-htmlbars-inline-precompile": "0.3.6",
+    "ember-cli-frost-blueprints": "^1.0.0",
+    "ember-cli-htmlbars-inline-precompile": "0.3.12",
```
quincyle: Re-run the build that was successful before, with no code change, and got build error from latest `ember-cli-htmlbars-inline-precompile` that just released. Pin it to 0.3.12 until it gets fixed.

config/ember-try.js
```diff
@@ -1,122 +1,8 @@
...
-        dependencies: {
-          'ember': '~2.7.0'
-        },
-        resolutions: {
-          'ember': '~2.7.0'
-        }
-      }
-    },
-    {
       name: 'ember-2-8',
```
notmessenger: This entry is not correct.  Make sure this file is identical to all the other versions of this file used in related upgrades of other repos

config/ember-try.js
```diff
@@ -125,16 +11,26 @@ module.exports = {
...
-      name: 'ember-2-9',
+      name: 'ember-lts-2.12',
       bower: {
         dependencies: {
-          'ember': '~2.9.0'
+          'ember': 'components/ember#lts-2-12'
         },
         resolutions: {
-          'ember': '~2.9.0'
+          'ember': 'lts-2-12'
```
juwara0: This can be updated to:

```
{
      name: 'ember-lts-2.12',
      npm: {
        devDependencies: {
          'ember-source': '~2.12.0'
        }
      }
    },
```

config/ember-try.js
juwara0: We had to do run on `ember-2-8` instead of `ember-lts-2.8` on this repo because there was a difference between those two releases that was causing failures in the build.

config/ember-try.js
quincyle: Missed that one. Thanks for the catch.

juwara0:
Approved

package.json
juwara0: This is most likely related to what is reported here: https://github.com/ciena-frost/ember-frost-core/issues/488

PR #466, opened 2017-09-12
Adds object/form renderer alongside much needed fixes

sophypal:

# CHANGELOG

* **Added** object/form renderer for rendering forms dynamically. Custom views can now reference an object type field directly without needing to expand the view definition further.
* **Added** plugins property (utilized by the form renderer) which is responsible for generating bunsen meta data.
* **Added** much needed test against the regressions found that degraded the performance.
* **Updated** refocus logic to focus on init (if applicable) instead of on every change.
* **Updated** the redux workflow to be more compatible with the run-loop. It should prevent issues with multiple synchronous updates causing redundant `onChange` actions being triggered which improves performance.
* **Fixed** performance issue with `when` renderer causing multiple `onChange` events on `init`.
* **Fixed** performance issue with value propagation due to a bug in the view precomputation and dependency diffing.
* **Removed** useless internal hooks. This does not alter the hook names of public interfaces but removes the duplicates used through the intermediate DOM nodes used to render a form. There are still duplicate hooks applied to the form inputs but no change has been made to those to avoid a major release.


sophypal:
Trying to resolve testing issue with the geolocation renderer since it seems it happens only in travis.

sophypal:
Tests resolved finally. Sigh!

gknoy:
@sophypal This looks pretty good. Thank you for explaining the usefulness of the new form renderer, and I like that using it as the default renderer for `object` fields means that the templates are simpler.

cstolli:
@spal I did look through this in some detail.  I don't see anything that jumps out at me but I'm not sure I would.  Test coverage is good, some nice improvements in async handling I see.  I say merge it if everything is passing.

sophypal:
👍

PR #467, opened 2017-09-13
Fix error when unsetting model/view

sophypal:

# CHANGELOG

* **Added** extra protection around setting of the model/views so that unsetting it after init wouldn't throw an exception.


notmessenger:
👍

PR #468, opened 2017-09-13
Add (breaking change) support for optional date pickers

gknoy:

### Breaking change:
The `date` renderer no longer defaults to the current date, so that we can present optional fields which do not have a value yet chosen, without mistakenly persisting today's date as a selection.

One can default to the current date by providing `defaultToCurrentDate: true` in the renderer options of the cell config.

 ### Usage
Default to `undefined`:
```javascript
// bunsen-view.js
{
  model: 'foo',
  renderer: {
    name: 'date'
  }
},
```
Default to the current date:
```javascript
// bunsen-view.js
{
  model: 'bar',
  renderer: {
    defaultToCurrentDate: true,
    name: 'date'
  }
}
```

### Example (from demo)
screen shot 2017-09-13 at 3 17 56 pm

 # Changelog
- Date widget no longer defaults to current date.  Specifying `defaultToCurrentDate: true` in the renderer options will re-enable this behavior.
- Use ember-frost-date-picker@7.3.0
- Add two different date renderers to demo

### DIFF
--- a/addon/components/inputs/date.js
+++ b/addon/components/inputs/date.js
@@ -5,6 +5,10 @@ import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-d
 
 const DATE_FORMAT = 'YYYY-MM-DD'
 
+export const deps = {
+  moment
+}
+
 export default AbstractInput.extend({
   // == Component Properties ===================================================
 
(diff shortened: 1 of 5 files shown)


addon/components/inputs/date.js
```diff
@@ -5,6 +5,10 @@ import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-d
 
 const DATE_FORMAT = 'YYYY-MM-DD'
 
+export const deps = {
```
sophypal: Interesting. I see you did this to make it stubbable.

sophypal:
👍

PR #475, opened 2017-09-27
Fix a crash on adding to array in complex models

TheAxeMan301:

- [ x] #patch# - backwards-compatible bug fix

# CHANGELOG
Prevent a crash that occurs on adding an array item in a form when the model has an array of objects with an array.

### DIFF
--- a/addon/components/array-container.js
+++ b/addon/components/array-container.js
@@ -137,7 +137,7 @@ export default Component.extend(HookMixin, PropTypeMixin, {
   @computed('bunsenModel.maxItems', 'value', 'bunsenId')
   maxItemsReached (maxItems, value, bunsenId) {
     if (value && maxItems) {
-      return get(value, bunsenId).length >= maxItems
+      return get(value, bunsenId + '.length') >= maxItems
     }
     return false
   },


TheAxeMan301:
If this simple and obvious one-line change doesn't pass I am probably not going to put any more work into this...

addon/components/array-container.js
```diff
@@ -137,7 +137,7 @@ export default Component.extend(HookMixin, PropTypeMixin, {
   @computed('bunsenModel.maxItems', 'value', 'bunsenId')
   maxItemsReached (maxItems, value, bunsenId) {
     if (value && maxItems) {
-      return get(value, bunsenId).length >= maxItems
+      return get(value, bunsenId + '.length') >= maxItems
```
notmessenger: Use templates literals rather than string concatenation.

notmessenger:
@TheAxeMan301

This "simple and obvious one-line change" has no explanation for what scenario it is addressing and no tests to ensure that the scenario does not re-occur.  Once that information is known it may be the case that this is not a "patch" but rather a "minor" due to the behavior for a consumer being changed in an unexpected way.

Additionally the build is failing.

TheAxeMan301:
Sorry, I didn't have much time but didn't want to forget about this fix. As it says in the PR, the problem happens when the model is nested with array of objects and a key in that object is an array. The fix is narrow and safe because it just changes a javascript access to an ember access that doesn't throw when the path isn't there.

A more thorough explanation of the bug: When you add an item to the outer array on the form the model adds an empty object as a placeholder. Then when it renders the form for the nested object that computed property gets checked to see if there should be an add button for the child array. Since the object is empty that get returns undefined and accessing `length` throws an exception that crashes Ember. The empty object is only a temporary state but that exception brings everything down.

Even if there was an issue with my fix, running the full test suite on the master tip currently fails locally for me. It takes over 10 minutes to do so and I have no idea why it doesn't work. If no one is able to figure things out then I can just use my fork and be on my way. I am not familiar with the test process and am not really interested in figuring it out.

cstolli:
@TheAxeMan301 we don't entertain pull requests from contributors that are unwilling to play by the rules.

PR #476, opened 2017-09-30
Fix race condition on form schemaLoaded cp and retrigger validation

sophypal:

# CHANGELOG

* **Fixes** race condition causing the `isSchemaLoaded` in the `form` renderer not to fire during init
* **Fixes** validation state that was caused by stale validation data. Calling `triggerValidation` on the input form's validation handler and when component validators are removed was necessary to sync up the proper validation state.


sophypal:
👍

PR #496, opened 2017-12-07
Fixed queryForCurrentValue with array values

TheOtherDude:

We have been using this branch in our app for months now, but I hesitated to make a PR because I didn't think it was clean enough, but we can just make it cleaner together if you guys have any ideas.

# CHANGELOG

* Fixed queryForCurrentValue with array values

### DIFF
--- a/addon/list-utils.js
+++ b/addon/list-utils.js
@@ -146,7 +146,7 @@ export function getItemsFromEmberData ({value, modelDef, data, bunsenId, store, 
   const valueAsId = get(value, bunsenId)
   let arrayValues
   if (isArray(valueAsId) && queryForCurrentValue) {
-    arrayValues = valueAsId.asMutable()  // frost-multi-select does not support immutable values
+    arrayValues = valueAsId.asMutable() // frost-multi-select does not support immutable values
     arrayValues = RSVP.Promise.all(arrayValues.map((id) => store.findRecord(modelType, id)))
       .catch((err) => {
         Logger.log(`Error fetching ${modelType}`, err)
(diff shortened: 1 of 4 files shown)


addon/list-utils.js
```diff
@@ -142,7 +142,12 @@ export function getItemsFromEmberData ({value, modelDef, data, bunsenId, store,
   const modelType = modelDef.modelType || 'resources'
   const {labelAttribute, queryForCurrentValue, valueAttribute} = modelDef
   const valueAsId = get(value, bunsenId)
-  const actuallyFindCurrentValue = keepCurrentValue && queryForCurrentValue && valueAsId !== undefined
+  let arrayValues
+  if (typeOf(valueAsId) === 'array' && queryForCurrentValue) {
```
sophypal: `Array.isArray`

addon/list-utils.js
```diff
@@ -142,7 +142,12 @@ export function getItemsFromEmberData ({value, modelDef, data, bunsenId, store,
...
+    arrayValues = valueAsId.asMutable()
+    arrayValues = RSVP.Promise.all(arrayValues.map((id) => store.findRecord(modelType, id)))
```
sophypal: No need to call `asMutable` if you're not going to mutate it. You might consider just inlining it with the `map`

sophypal: Help me understand the use case. Is this for `multi-select`? Also, needs tests.

sophypal:
Help me understand the use case. Is this for multi-select? Also, needs tests.

TheOtherDude:
Yes, this is for multi-selects. 

Also, note that, unlike with single-select, currently selected records as always included in the drop down. I find this to be a useful feature because you can easily view and uncheck selections, but I wanted to bring it up because I could see others regarding it as a bug.

addon/list-utils.js
TheOtherDude: I added a comment explaining why I'm calling `asMutable()`

sophypal:
Thanks @TheOtherDude

sophypal:
👍

PR #498, opened 2017-12-19
fix required labels to use the showRequiredLabel prop

gkchestertron:

# CHANGELOG
* make date, datetime, when, and section show required label based on showRequiredLabel prop

### DIFF
--- a/addon/templates/components/frost-bunsen-section.hbs
+++ b/addon/templates/components/frost-bunsen-section.hbs
@@ -18,7 +18,7 @@
       {{title}}
     </h3>
   {{/if}}
-  {{#if required}}
+  {{#if showRequiredLabel}}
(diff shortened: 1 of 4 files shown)


addon/templates/components/frost-bunsen-section.hbs
```diff
@@ -18,7 +18,7 @@
       {{title}}
     </h3>
   {{/if}}
-  {{#if required}}
+  {{#if showRequiredLabel}}
```
sophypal: I don't think `frost-bunsen-section` has the `showRequiredLabel` computed property does it?

gkchestertron: No - you are right.

sophypal:
👍

PR #499, opened 2017-12-19
Fix type checking for null

sophypal:

# CHANGELOG

* Fixed an issue when parsing json with `null` that was caused by incorrectly type checking for `null`.

### DIFF
--- a/addon/utils.js
+++ b/addon/utils.js
@@ -451,24 +451,16 @@ function maybeMerge (first, second) {
  * @returns {any} The value with any _internal properties removed
  */
 export function removeInternalValues (val) {
-  if (['boolean', 'number', 'string', 'undefined', 'null'].includes(typeof val)) {
+  const valueTypes = ['boolean', 'number', 'string']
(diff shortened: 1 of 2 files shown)


addon/utils.js
```diff
@@ -451,24 +451,16 @@ function maybeMerge (first, second) {
  * @returns {any} The value with any _internal properties removed
  */
 export function removeInternalValues (val) {
-  if (['boolean', 'number', 'string', 'undefined', 'null'].includes(typeof val)) {
+  const valueTypes = ['boolean', 'number', 'string']
+  const emptyTypes = [undefined, null]
+
+  if (emptyTypes.includes(val) || valueTypes.includes(typeof val)) {
```
gkchestertron: might be cleaner/faster to say:
```
if (typeof val !== 'object' || val === null) {
```

sophypal: k, like that better.

sophypal:
👍

PR #507, opened 2018-02-01
Array should provide customizable `Add` button text, as acronyms are incorrectly translated

### LINKED ISSUE #506: Array should provide customizable `Add` button text, as acronyms are incorrectly translated
see: http://ciena-frost.github.io/ember-frost-bunsen/#/examples?model=array&view=array

 ```
"children": [
        {
          "model": "name",
          "extends": "name"
        },
        {
          "model": "addresses",
          "arrayOptions": {
            "itemCell": {
              "extends": "addr",
              "label": "Address"
            }
          }
        }
      ]
```

I cannot find a way to customize the array `Add` button text, here are some of the issues:

array-add-label-issue2

array-add-label-issue1

notmessenger:

Closes #506

# CHANGELOG
* Section heading text now respects the `label` property.  When set it will return the value of the property with the first character capitalized, whereas before it would use the exact value provided.
* "Add" button in the array container now respects the `label` property.  When set it will return a singularized version of the text without affecting the case, whereas before it would lowercase the text.


juwara0:
👍

juwara0:
👍

PR #522, opened 2018-04-30
adding autocomplete support to bunsen

### LINKED ISSUE #521: Add autocomplete support into bunsen
# FEATURE REQUEST / CODE CHANGE

## Summary
frost-autocomplete was added into ember-frost-core and needs support in bunsen

## Motivation
This lets the user start typing right away to find what they want. They will not see anything until they start typing though

## Detailed design
The autocomplete will extend the functionality that frost-select does, as they have the same inputs. This will allow them to be interchangeable

## How we teach this
autocomplete will be added into the dummy app.

helrac:
# Overview

## Summary
This implemented the autocomplete renderer into bunsen. Autocomplete has also been added into the dummy app.

* Closes #521 

## Screenshots or recordings
Please provide screenshots or recordings if this PR is modifying the visual UI or UX.

# Semver

- [ ] #none#
- [ ] #patch#
- [x] #minor#
- [ ] #major#

# CHANGELOG

* **Added** autocomplete renderer to bunsen (interchangeable with select)


tests/dummy/app/pods/view/renderers/documentation/autocomplete.md
```diff
@@ -0,0 +1,259 @@
+## autocomplete
+
+This renderer provides an autocomplete input which allows one option to be selected.
```
sophypal: Expand on he differences between `select` and `autocomplete` please.

sophypal: Also, the rest of the docs is a dup to `select`. I'd hate to have to maintain 2 separate docs if the implementation for one is altered. Perhaps provide a link and mention that the API is the same as `select`.

tests/dummy/app/pods/view/renderers/models/autocomplete.js
```diff
@@ -0,0 +1,14 @@
+export default {
+  properties: {
+    cookies: {
+      enum: [
```
sophypal: Maybe using an `enum` isn't the best choice for demo. Since you're trying to show that `autocomplete` is a little different from select. Using an `enum` the value is validated again the `enum`. For `autocomplete` I believe free text is allowed which would fail validation if that text wasn't in`enum`.

sophypal: In fact, you might want to provide some protection against using `enum` for `autocomplete`. Do you agree?

addon/components/inputs/autocomplete.js
```diff
@@ -0,0 +1,43 @@
...
+  // == Actions ================================================================
+
+  actions: {
+    /**
+     * perform a filter on the widget
+     * @param  {String} filter the filter text
+     */
+    filterOptions (filter) {
+      const value = this.get('formValue')
+      run.debounce(this, this.runUpdateItems, {value, filter, keepCurrentValue: false}, DEBOUNCE_INTERVAL)
```
sophypal: Why does the `autocomplete` get the `debounce` but `select` doesn't?

tests/dummy/app/pods/view/renderers/documentation/autocomplete.md
helrac: ya, i'll reference select. I didn't like duping it anyways

tests/dummy/app/pods/view/renderers/models/autocomplete.js
helrac: Well, it's just a different style of dropdown, i see no reason to block using enum. It's up to the end user. The example under examples for autocomplete/autocomplete uses a query. I could switch that example to use a query

addon/components/inputs/autocomplete.js
helrac: Well, I didn't want to change the functionality of select, I could give it to select but give it a debounce of 0 unless otherwise overridden if you'd like

tests/dummy/app/pods/view/renderers/models/autocomplete.js
sophypal: It's not just style that's different though. With `select` you have to choose from the list of available options. With `autocomplete` you don't have to. This means the value can optionally come from the dropdown list and that behavior breaks the meaning of `enum`.

So I can understand not wanting to alter the behavior to intentionally block `enum` but at the very least, lets not encourage people to use it with `enum`.

helrac: yeah, in the documentation I will list what scenario is ideal for autocomplete. IE, wanting the user to type to search for options

addon/components/inputs/select.js
```diff
@@ -551,6 +551,7 @@ export default AbstractInput.extend({
      */
     filterOptions (filter) {
       const value = this.get('formValue')
+      console.log('filtering for ' + filter)
```
sophypal: Remove

package-lock.json
```diff
@@ -952,14 +953,14 @@
     },
     "autoprefixer": {
       "version": "7.2.6",
-      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-7.2.6.tgz",
-      "integrity": "sha512-Iq8TRIB+/9eQ8rbGhcP7ct5cYb/3qjNYAR2SnzLCEcwF6rvVOax8+9+fccgXk4bEhQGjOZd5TLhsksmAdsbGqQ==",
+      "resolved": "https://artifactory.ciena.com/api/npm/blueplanet-npm/autoprefixer/-/autoprefixer-7.2.6.tgz?dl=https://registry.npmjs.org/autoprefixer/-/autoprefixer-7.2.6.tgz",
```
juwara0: This is incorrect. Since this is a public repo you do not want it to point at `artifactory`. To fix this you will need to remove the setting in your dev environment that is setting the `registery` to `artifactory` for everything and instead place a `.npmrc` file with that `registry` setting in your directories for private repos only.

package.json
```diff
@@ -87,7 +87,7 @@
     "ember-cli-svgstore": "github:ciena-blueplanet/ember-cli-svgstore#977df1cf58ae43b1d98a591573c3e06947744321",
     "ember-computed-decorators": "0.3.0",
     "ember-concurrency": "0.7.19",
-    "ember-frost-core": "^8.0.0",
+    "ember-frost-core": "^8.2.2",
```
juwara0: This change should be reverted. It is not necessary to update the version of `ember-frost-core` here since it will already pick up the minor due to the `^` float.

PR #540, opened 2018-07-18
Fix autocomplete not having filter blank after selecting an item, and fix 

EWhite613:
# Overview

## Does this PR close an existing issue?
No PR should be opened without opening an issue first.  Any change needs to be discussed before proceeding.

## Summary

## Issue Number(s)
Which issue(s) does this PR address?

Put `Closes #XXXX` below to auto-close the issue that this PR addresses:

* Closes #

## Screenshots or recordings
Please provide screenshots or recordings if this PR is modifying the visual UI or UX.

## Checklist
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] I have evaluated if the _README.md_ documentation needs to be updated
* [ ] I have evaluated if the _/tests/dummy/_ app needs to be modified
* [ ] I have verified that lint and tests pass locally with my changes

# Semver

- [ ] #none#
- [x] #patch#
- [ ] #minor#
- [ ] #major#

Examples:
* **NONE**
  * _README.md_ changes
  * test additions
  * changes to files that are not used by a consuming application (_.travis.yml_, _.gitignore_, etc)
* **PATCH**
  * backwards-compatible bug fix
    * nothing about how to use the code has changed
    * nothing about the outcome of the code has changed (though it likely corrected it)
  * changes to demo app (_/tests/dummy/_)
* **MINOR**
  * adding functionality in a backwards-compatible manner
    * nothing about how used to use the code has changed but using it in a new way will do new things
    * nothing about the outcome of the code has changed without having to first use it in a new way
    * addition of new CSS selectors
    * addition of new `ember-hook` selectors
* **MAJOR**
  * incompatible API change
    * using the code how used to will cease working
    * using the code how used to will have a different outcome
    * any changes to CSS selector names
    * any removal of CSS selectors
    * any changes to `ember-hook` selectors
    * possibly changes to test helpers (depends on the changes made)
  * any changes to the **_dependencies_** entry in the _package.json_ file

# CHANGELOG
* **Fix**  autocomplete not being able to backspace to blank filter when have already selected

### DIFF
--- a/addon/components/inputs/autocomplete.js
+++ b/addon/components/inputs/autocomplete.js
@@ -1,7 +1,7 @@
 import SelectInput from './select'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-autocomplete'
 import Ember from 'ember'
-const {A, isEmpty} = Ember
+const {A, get, getWithDefault, isEmpty, observer} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 
 export default SelectInput.extend({
@@ -16,10 +16,11 @@ export default SelectInput.extend({
   getDefaultProps () {
     return {
       ignoreEmptyFilterSearch: true,
-      filter: ''
+      filter: '',
+      _focused: false,
+      _isTyping: false
     }
   },
-
   @readOnly
   @computed('isAsyncGet', 'updateItems.isRunning', '_emptyFilter')
   asyncLoading (isAsyncGet, isUpdateItemsRunning, emptyFilter) {
@@ -37,10 +38,38 @@ export default SelectInput.extend({
   @computed('value', 'options.[]')
   selectedItemWithLabel (value, options) {
     if (typeof value === 'string' && options) {
-      return A(options).findBy('value', value) || value
+      return this._findSelectedItemGivenValue(value, options) || value
     }
 
     return value
+  },
+
+  /**
+   * Return selectedItem given value chosen
+   * @param {String} value - value of selected item
+   * @param {Array} data - array of items displayed in autocomplete
+   * @returns {Object} selected item
+   */
+  _findSelectedItemGivenValue (value, data) {
+    if (typeof value === 'string' && data) {
+      return A(data).findBy('value', value) || value
+    }
+
+    return value
+  },
+  /**
+   * Return selectedItem given value chosen
+   * @param {String} value - value of selected item
+   * @param {Array} data - array of items displayed in autocomplete
+   * @returns {String} selected item label
+   */
+  _findSelectedItemLabelGivenValue (value, data) {
+    let label = getWithDefault(value, 'label', '')
+    if (typeof selectedItem === 'string') {
+      const foundItem = this._findSelectedItemGivenValue(value, this.get('options'))
+      label = get(foundItem, 'label')
+    }
+    return label
   },
   /**
    * This should be overriden by inherited inputs to convert the value to the appropriate format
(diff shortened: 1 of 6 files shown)


addon/components/inputs/autocomplete.js
```diff
@@ -16,10 +16,29 @@ export default SelectInput.extend({
...
-      filter: ''
+      filter: '',
+      _focused: false,
+      _isTyping: false
     }
   },
 
+  // == Observers ====================================================
+
+  observeSelectedItemLavelChange: observer('value', 'options.[]', '_isTyping', 'filter', function () {
```
sophypal: Lavel?

addon/components/inputs/autocomplete.js
```diff
@@ -16,10 +16,29 @@ export default SelectInput.extend({
...
+    const {
+      _isTyping: isTyping,
+      filter,
+      options,
+      value
+    } = this.getProperties('value', 'options', '_isTyping', 'filter')
+    const selectedItem = this._findSelectedItemGivenValue(value, options)
+    if (!isEmpty(selectedItem)) {
+      const label = get(selectedItem, 'label')
+      if (!isEmpty(label) && isEmpty(filter) && !isTyping) this.set('filter', label)
```
sophypal: So I'm having a tough time understanding what the fix is. Do you mind posting screenshots/vids of the before and after. From what I can understand from these changes, it looks like the `filter` was empty previously and you're setting it the the selected item label when user is not typing? Is that correct?

addon/components/inputs/autocomplete.js
```diff
@@ -50,14 +89,36 @@ export default SelectInput.extend({
...
+  // == Actions ===============================================================
   actions: {
-    checkIfEmptyfilter (filterValue) {
-      this.set('filter', filterValue)
+    onInput (filterValue) {
+      const focused = this.get('_focused')
+
+      this.setProperties({
+        filter: filterValue,
+        _isTyping: focused
```
sophypal: Interested to know why it's necessary to keep typing state. Is there no alternative?

sophypal: I probably just need to get a better understanding of the bug you're fixing.

EWhite613: Ya I'll add some screenshots/gifs tomorrow to better explain the problem

addon/components/inputs/autocomplete.js
EWhite613: Well the original problem when you have a selectedItem and backspace to an empty filter it would.


So if I can keep track of typing I know that I shouldn't set the filter when the user is just backspacing all the way

addon/components/inputs/autocomplete.js
EWhite613: Well the original problem when you have a selectedItem and backspace to an empty filter it would.


So if I can keep track of typing I know that I shouldn't set the filter when the user is just backspacing all the way

addon/components/inputs/autocomplete.js
sophypal: Interesting. Originally I was thinking that if a user started typing, it meant user intention was to set the value (even with backspace to clear) so it shouldn't need to set the filter back to the original. But `autocomplete` isn't like Google's auto suggest. A value is only made with a selection and filtering the list shouldn't clear the value.

This was easier to handle in `select` because there was clear separation between the filtering input and the value's input element. In `select` typing anything would update the filter to whatever the user typed, regardless if it was empty. If the user didn't like the choices, he could close the dropdown and get back his original value.

Would it be better to keep track of the dropdown states instead of the typing? If the dropdown is open, don't override the filter ever; just keep it at whatever the user typed. Hmm. that would mean you'd also have to fix the automatic closing of the dropdown when they backspace to clear. Man, this component is getting really complex.

Up to you I guess. If I had a hard time understanding this scenario, others will and it'll be hard to maintain. Perhaps throw in some comments explaining this.

EWhite613: The problem, in this case, is I don't have knowledge on whether drop-down is open or not. That's in `frost-autocomplete`, and since we need to use `onInput` for querying/etc we can't really have `frost-autocomplete` control filter for us (because `frost-autocomplete` assumes if you use `onInput` you will control the `filter`, and shouldn't set it.


I think I made it much more readable now by removing the observer logic, I'll add some comments on `_isTyping` to make it easier to read

EWhite613: Looks like drop-down state is also closed when you have an empty filter. So I'd run into the same issue

sophypal: Thanks, removing the observer business definitely makes this more readable.

PR #543, opened 2018-07-24
Allow blank value in datetime renderer

jscharf:
# Overview

## Does this PR close an existing issue?
No PR should be opened without opening an issue first.  Any change needs to be discussed before proceeding.

## Summary
This PR allows for developers to configure a datetime renderer to allow a blank value (`undefined`)

## Issue Number(s)
Which issue(s) does this PR address?

Put `Closes #XXXX` below to auto-close the issue that this PR addresses:

* Closes #

## Screenshots or recordings
An example of what this new API looks like in a Bunsen view context
```javascript
{
  "cells": [
    {
      "model": "foo",
      "renderer": {
        "name": "datetime",
        "options": {
          "defaultToCurrentDateTime": false
        }
      }
    }
  ],
  "type": "form",
  "version": "2.0"
}
```

## Checklist
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] I have evaluated if the _README.md_ documentation needs to be updated
* [x] I have evaluated if the _/tests/dummy/_ app needs to be modified
* [x] I have verified that lint and tests pass locally with my changes

# Semver

- [ ] #none#
- [ ] #patch#
- [x] #minor#
- [ ] #major#

Examples:
* **NONE**
  * _README.md_ changes
  * test additions
  * changes to files that are not used by a consuming application (_.travis.yml_, _.gitignore_, etc)
* **PATCH**
  * backwards-compatible bug fix
    * nothing about how to use the code has changed
    * nothing about the outcome of the code has changed (though it likely corrected it)
  * changes to demo app (_/tests/dummy/_)
* **MINOR**
  * adding functionality in a backwards-compatible manner
    * nothing about how used to use the code has changed but using it in a new way will do new things
    * nothing about the outcome of the code has changed without having to first use it in a new way
    * addition of new CSS selectors
    * addition of new `ember-hook` selectors
* **MAJOR**
  * incompatible API change
    * using the code how used to will cease working
    * using the code how used to will have a different outcome
    * any changes to CSS selector names
    * any removal of CSS selectors
    * any changes to `ember-hook` selectors
    * possibly changes to test helpers (depends on the changes made)
  * any changes to the **_dependencies_** entry in the _package.json_ file

# CHANGELOG

* Allow developer to configure a datetime renderer to display a blank value

### DIFF
--- a/addon/components/inputs/datetime.js
+++ b/addon/components/inputs/datetime.js
@@ -1,11 +1,8 @@
-import Ember from 'ember'
 import computed, {readOnly} from 'ember-computed-decorators'
 import moment from 'moment'
 
 import AbstractInput from './abstract-input'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-datetime'
-
-const {isEmpty} = Ember
 
 const DATE_FORMAT = 'YYYY-MM-DD'
 const TIME_FORMAT = 'HH:mm:ss'
@@ -27,10 +24,7 @@ export default AbstractInput.extend({
 
   @readOnly
   @computed('cellConfig.renderer.options.defaultToCurrentDateTime')
-  defaultToCurrentDateTime (defaultToCurrentDateTime) {
-    if (isEmpty(defaultToCurrentDateTime)) {
-      return true
-    }
+  defaultToCurrentDateTime (defaultToCurrentDateTime = true) {
     return Boolean(defaultToCurrentDateTime)
   },
 
--- a/addon/components/inputs/datetime.js
+++ b/addon/components/inputs/datetime.js
@@ -1,8 +1,11 @@
+import Ember from 'ember'
 import computed, {readOnly} from 'ember-computed-decorators'
 import moment from 'moment'
 
 import AbstractInput from './abstract-input'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-datetime'
+
+const {isEmpty} = Ember
 
 const DATE_FORMAT = 'YYYY-MM-DD'
 const TIME_FORMAT = 'HH:mm:ss'
@@ -23,14 +26,11 @@ export default AbstractInput.extend({
   // == Computed Properties ====================================================
 
   @readOnly
-  @computed('cellConfig.renderer.options.allowBlank')
-  allowBlank (allowBlank) {
-    return Boolean(allowBlank)
-  },
-
-  @readOnly
   @computed('cellConfig.renderer.options.defaultToCurrentDateTime')
   defaultToCurrentDateTime (defaultToCurrentDateTime) {
+    if (isEmpty(defaultToCurrentDateTime)) {
+      return true
+    }
     return Boolean(defaultToCurrentDateTime)
   },
 
@@ -49,9 +49,9 @@ export default AbstractInput.extend({
   },
 
   @readOnly
-  @computed('allowBlank', 'defaultToCurrentDateTime', 'value')
-  currentValue (allowBlank, defaultToCurrentDateTime, value) {
-    if ((allowBlank && !defaultToCurrentDateTime) || value) {
+  @computed('defaultToCurrentDateTime', 'value')
+  currentValue (defaultToCurrentDateTime, value) {
+    if (!defaultToCurrentDateTime || value) {
       return value
     }
     return moment().format(DATE_TIME_FORMAT)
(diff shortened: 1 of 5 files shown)


addon/components/inputs/datetime.js
```diff
@@ -22,6 +22,18 @@ export default AbstractInput.extend({
 
   // == Computed Properties ====================================================
 
+  @readOnly
+  @computed('cellConfig.renderer.options.allowBlank')
+  allowBlank (allowBlank) {
```
sophypal: Is `allowBlank` really necessary? One would assume that if a default wasn't automatically set, it would be blank. I'm trying to keep this consistent with what was done to the `date` renderer which only added `defaultToCurrentDate` to the list of options.

jscharf: My thinking behind `allowBlank` was that the current behavior is to default to the current date, and not allow a blank value. I wanted to preserve the default behavior as much as possible, so by adding `allowBlank` I thought that existing code would be unaffected since none of it would have that option. Then I added `defaultToCurrentDateTime` to be used in conjunction so that people could still use the current default if they wanted to. 

Now that you mention it though, if there was just one setting `defaultToCurrentDateTime` and it was true by default (for backward compatibility), then people could set it to false if they wanted the field to be blank. If you think just having one property instead of two is better in this case, I will make the change.

sophypal: I do. It's also much simpler to keep track of and consistent with changes to other date components.

addon/components/inputs/datetime.js
```diff
@@ -22,6 +25,15 @@ export default AbstractInput.extend({
 
   // == Computed Properties ====================================================
 
+  @readOnly
+  @computed('cellConfig.renderer.options.defaultToCurrentDateTime')
+  defaultToCurrentDateTime (defaultToCurrentDateTime) {
```
sophypal: Can't you just set the default as a defaulted parameter.

addon/components/inputs/datetime.js
```diff
@@ -22,6 +25,15 @@ export default AbstractInput.extend({
...
+    if (isEmpty(defaultToCurrentDateTime)) {
+      return true
+    }
+    return Boolean(defaultToCurrentDateTime)
```
sophypal: Are you allowing strings and numbers?

tests/dummy/app/pods/view/renderers/documentation/datetime.md
```diff
@@ -15,4 +15,57 @@ by passing disabled = true to form but not through date-time-picker.
...
+  "model": "foo",
+  "renderer": {
+    "name": "datetime",
+    "options": {
+    }
+  }
+}
+```
+
+##### allowBlank
```
sophypal: Remove

tests/dummy/app/pods/view/renderers/documentation/datetime.md
```diff
@@ -15,4 +15,57 @@ by passing disabled = true to form but not through date-time-picker.
...
+    "options": {
+      "allowBlank": true
+    }
+  }
+}
+```
+
+##### defaultToCurrentDateTime
+
+The `defaultToCurrentDateTime` option can be used in conjunction with `allowBlank` to preserve the default functionality
```
sophypal: Update without allowBlank

addon/components/inputs/datetime.js
jscharf: I am but only because I not familiar with the codebase enough to know how to properly restrict the values to only `false` and `true`. I can add PropTypes to the file to indicate that it's supposed to be a boolean, and look into validation to ensure that no non-boolean values are accepted.

addon/components/inputs/datetime.js
jscharf: I'm not sure exactly what you mean here, can you elaborate?

addon/components/inputs/datetime.js
sophypal: Personally, I find it's enough to document the types either in code or in the renderers documentation. If you want to go the validation route, you'd have to make an accompanying change in the UISchema2 validation.

addon/components/inputs/datetime.js
jscharf: Oh, yeah that's cleaner, I'll do that.

addon/components/inputs/datetime.js
jscharf: Well if it's up to me, I'd rather trust developers to use the tool properly. Especially if the tool is documented, and then the developer misuses it, my philosophy is that's on them. I initially looked at the date renderer and didn't see any special type handling in there for the `defaultToCurrentDate` property, so I didn't feel the need to add any here either. I added the `Boolean` cast in a small effort to maintain composure just in case a non-boolean value was given.

PR #544, opened 2018-07-30
Add timezone to datetime and when renderers

anhuynh:
# Overview
Add timezone into `datetime` and `when` renderers

## Summary

## Screenshots or recordings
screen shot 2018-07-30 at 10 02 23 am
screen shot 2018-07-30 at 10 24 47 am

## Checklist
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] I have evaluated if the _README.md_ documentation needs to be updated
* [x] I have evaluated if the _/tests/dummy/_ app needs to be modified
* [ ] I have verified that lint and tests pass locally with my changes

# Semver

- [ ] #none#
- [ ] #patch#
- [x] #minor#
- [ ] #major#

# CHANGELOG

* Add timezone options to `datetime` and `when` renderers

### DIFF
--- a/addon/components/inputs/abstract-input.js
+++ b/addon/components/inputs/abstract-input.js
@@ -1,7 +1,7 @@
 import {getCellDefaults, utils} from 'bunsen-core'
 const {getLabel, parseVariables} = utils
 import Ember from 'ember'
-const {Component, Logger, get, merge, typeOf} = Ember
+const {Component, Logger, get, getWithDefault, merge, typeOf} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 import {HookMixin} from 'ember-hook'
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
@@ -154,6 +154,12 @@ export default Component.extend(HookMixin, PropTypeMixin, {
     )
 
     return required && valueEmpty
+  },
+
+  @readOnly
+  @computed('cellConfig')
+  alwaysShowRequiredLabel (cellConfig) {
+    return getWithDefault(cellConfig, 'renderer.alwaysShowRequiredLabel', false)
   },
 
   // == Functions ==============================================================
--- a/addon/components/inputs/abstract-input.js
+++ b/addon/components/inputs/abstract-input.js
@@ -1,7 +1,7 @@
 import {getCellDefaults, utils} from 'bunsen-core'
 const {getLabel, parseVariables} = utils
 import Ember from 'ember'
-const {Component, Logger, get, getWithDefault, merge, typeOf} = Ember
+const {Component, Logger, get, merge, typeOf} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 import {HookMixin} from 'ember-hook'
 import PropTypeMixin, {PropTypes} from 'ember-prop-types'
@@ -154,12 +154,6 @@ export default Component.extend(HookMixin, PropTypeMixin, {
     )
 
     return required && valueEmpty
-  },
-
-  @readOnly
-  @computed('cellConfig')
-  alwaysShowRequiredLabel (cellConfig) {
-    return getWithDefault(cellConfig, 'renderer.alwaysShowRequiredLabel', false)
   },
 
   // == Functions ==============================================================
--- a/addon/components/inputs/datetime.js
+++ b/addon/components/inputs/datetime.js
@@ -6,7 +6,7 @@ import AbstractInput from './abstract-input'
 import layout from 'ember-frost-bunsen/templates/components/frost-bunsen-input-datetime'
 import {DEFAULT_TIMEZONE, getFormattedTime, getMomentTimezone} from 'ember-frost-bunsen/timezone-utils'
 
-const {getWithDefault} = Ember
+const {getWithDefault, isNone, run} = Ember
 const DATE_FORMAT = 'YYYY-MM-DD'
 const TIME_FORMAT = 'HH:mm:ss'
 const DATE_TIME_FORMAT = 'YYYY-MM-DDTHH:mm:ssZ'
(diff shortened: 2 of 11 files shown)


addon/components/inputs/abstract-input.js
```diff
@@ -156,6 +156,12 @@ export default Component.extend(HookMixin, PropTypeMixin, {
     return required && valueEmpty
   },
 
+  @readOnly
+  @computed('cellConfig')
+  alwaysShowRequiredLabel (cellConfig) {
```
sophypal: Why was this necessary?

addon/components/inputs/datetime.js
```diff
@@ -43,12 +53,14 @@ export default AbstractInput.extend({
   },
 
   @readOnly
-  @computed('defaultToCurrentDateTime', 'value')
-  currentValue (defaultToCurrentDateTime, value) {
-    if (!defaultToCurrentDateTime || value) {
-      return value
+  @computed('defaultToCurrentDateTime', 'timezone', 'storedDateTimeValue')
```
sophypal: Why are you moving away from `value`? If consumers decide to pass in a value directly into the `formValue`, this renderer won't update with the correct value. It only seems to initialize with the correct value at initialization, not updates.

addon/components/inputs/datetime.js
```diff
@@ -57,8 +69,43 @@ export default AbstractInput.extend({
...
+     * case that the user selects away from the date-time picker radio button but then returns
+     * @param {Object} value - the users selected date/time moment() object
+     * @returns {undefined}
+     */
+    selectDate (value) {
+      const timezone = this.get('timezone')
+      const datetime = getMomentTimezone(value, timezone, true)
+      const formattedTime = getFormattedTime(datetime, this.get('dateTimeFormat'), timezone)
+
+      this.set('storedDateTimeValue', formattedTime)
```
sophypal: This internal state will have the tendency to get out of date with the actual form's value. In future versions of Bunsen, we won't be re-rendering components needlessly so you can't rely on `init` to keep it in sync.

addon/components/inputs/when.js
```diff
@@ -106,6 +108,19 @@ export default AbstractInput.extend({
...
 
+  @readOnly
+  @computed('cellConfig')
+  timezone (cellConfig) {
+    return getWithDefault(cellConfig, 'renderer.timezone', '')
+  },
+
+  @readOnly
+  @computed('storedDateTimeValue', 'timezone')
+  displayStoredDateTimeValue (storedDateTimeValue, timezone) {
```
sophypal: Hmm. I realize now the `storedDateTimeValue` is a artifact of the `when` renderer which you probably just copied. `when` renderer does suffer from the same problems I mentioned.

addon/components/inputs/abstract-input.js
anhuynh: I thought I needed that but actually don't so I'll remove it

addon/components/inputs/when.js
anhuynh: Ah yes, after closer inspection, I now see why the `storedDateTimeValue` is needed in the `when` renderer. I'll go back to using `value` in `datetime` but not sure if we can get rid of `storedDateTimeValue` in the `when` renderer

sophypal: Thanks. Stick with datetime for now. We'll address `when` at a future date.

addon/components/inputs/datetime.js
```diff
@@ -57,8 +69,47 @@ export default AbstractInput.extend({
...
+    this._super(...arguments)
+
+    const value = this.get('value')
+    const timezone = this.get('timezone')
+    let currentDateTime = getMomentTimezone(value, timezone)
+
+    if (!this.get('defaultToCurrentDateTime')) {
+      currentDateTime = value
+    } else {
+      currentDateTime = getFormattedTime(currentDateTime, this.get('dateTimeFormat'), timezone)
```
sophypal: If I'm reading this correctly, it says if `defaultToCurrentDateTime` is true, then it always overrides the value to the current datetime. I don't think that's what we want. We only want to default it when there is no value.

addon/components/inputs/datetime.js
```diff
@@ -57,8 +69,47 @@ export default AbstractInput.extend({
...
+    }
+
+    this.setProperties({
+      localTimezone: moment().format('Z')
+    })
+
+    if (!isNone(value)) {
```
sophypal: You should only call `onChange` if the value is different previously. For example, if the initial value was `null` or `undefined`, you'd want to call `onChange`.  If `value` is `none`, then you'd want to call `onChange`.

addon/templates/components/frost-bunsen-input-datetime.hbs
```diff
@@ -1,7 +1,7 @@
 {{#if (not cellConfig.hideLabel)}}
   <label class={{labelWrapperClassName}}>
     {{renderLabel}}
-    {{#if showRequiredLabel}}
+    {{#if (or showRequiredLabel alwaysShowRequiredLabel)}}
```
sophypal: You don't need this anymore

addon/components/inputs/datetime.js
anhuynh: Yeah, that makes sense. I think that this and the stuff you commented on above actually aren't needed. Not quite sure what I was thinking behind this

anhuynh:
@jscharf That looks pretty good. Would have to pass that to UX to see what they think though

addon/components/inputs/datetime.js
sophypal: That's quite alright. I think the idea was sane. Just needed a little more polish to make sure it worked properly. I would have been okay if you left it but corrected it. As it stands, the component does suffer from not setting the value when the user wants defaults applied. The `currentValue` is only presentational until the first date selection is made.

Leaving it out is fine as well since it's only returning to the same behavior as before.

jscharf:
@anhuynh thanks. I don't want to block this PR if it's good to go, so could you follow up with UX and then do another PR based on their feedback?

PR #545, opened 2018-07-31
Fix re-rendering issues on schema changes

sophypal:
#PATCH#

# CHANGELOG

* Used keys for `each` to prevent re-rendering when the cell or model hasn't changed
* Don't load temporary bunsen form on init, but rather wait until schema has been processed so Ember doesn't try and render the form twice
* Fixed tests associated with select and required as required changes no longer causes re-render

### DIFF
--- a/addon/components/detail.js
+++ b/addon/components/detail.js
@@ -15,10 +15,7 @@ const {
   validate
 } = actions
 
-const {getSubModel} = utils
-
 import Ember from 'ember'
-const {A, Component, Logger, RSVP, getOwner, isEmpty, isPresent, run, typeOf} = Ember
 import computed, {readOnly} from 'ember-computed-decorators'
 import {task, timeout} from 'ember-concurrency'
 import {HookMixin} from 'ember-hook'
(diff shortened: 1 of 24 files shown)


addon/components/detail.js
```diff
@@ -376,18 +381,23 @@ export default Component.extend(SpreadMixin, HookMixin, PropTypeMixin, {
...
       }
+
+      subModel = getSubModel(bunsenModel, cellConfig.model, cellConfig.dependsOn) || {}
     }
 
+    let cellKey = index
     if (!cellConfig.children || cellConfig.children.length === 0) {
       // only add bunsenIds to leaf nodes
       addMetaProperty(cellConfig, '__bunsenId__', bunsenId)
+      cellKey = JSON.stringify(cellConfig) + JSON.stringify(subModel)
```
sophypal: `JSON.stringify` is used as a hash to uniquely identify a sub schema from another. Testing has shown that arranging different layouts with the same views resulted in no re-rendering and outperforms using the cell index as the key. `JSON.stringify` is only used on leaf cells and the view config be should relatively small in comparison to the entire form schemas. model schemas on average is smaller than the view config so even if a large model is referenced, it's not enormous. With these constraints I was comfortable with the performance of `JSON.stringify` as it doesn't add much to the cost.

EWhite613: This will miss ordering though `JSON.stringify({a: '2', b: 2}) === JSON.stringify({b: 2, a: '2'}) // false`. What are the consequences of that, just losing efficiency?

sophypal: When the ordering is different, it usually means that the schema author had redefined the schema using conditions or some other means. In those cases, it's okay for the cell to be re-rendered.

sophypal: The re-rendering I'm trying to avoid is those cells that didn't undergo change.

PR #550, opened 2018-08-21
Fix re rendering

sophypal:
#PATCH#

# CHANGELOG

* Fix `package-lock`


sophypal:
Self merging to get demo updated

PR #551, opened 2018-08-22
Add loading indicator for frost-select when doing async filter

EWhite613:
# Overview

## Screenshots/Recordings
image

## Summary

# Semver

- [ ] #none#
- [ ] #patch#
- [x] #minor#
- [ ] #major#

# CHANGELOG

* **Feature**, add loading indicator for frost-select when doing async filter

### DIFF
--- a/addon/components/inputs/select.js
+++ b/addon/components/inputs/select.js
@@ -69,6 +69,14 @@ export default AbstractInput.extend({
   selectedOptions: [],
 
   // == Computed Properties ====================================================
+  @readOnly
+  @computed('isAsyncGet', 'updateItems.isRunning')
+  asyncLoading (isAsyncGet, isUpdateItemsRunning) {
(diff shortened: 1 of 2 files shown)


sophypal:
Looks good. Tests are failing.

EWhite613:
Hmmm very strange error with spread it seems will result in `Cannot set read-only property 'placeholder' on object: <dummy@component:frost-select::ember178008>`

`placeholder` isn't a readOnly property. 

When both
```
options=selectSpreadProperties
placeholder=placeholder
```
are present it will get the error. But if you remove `placeholder=placeholder` it won't error anymore, somehow the new `@readOnly` manages to be placed on `placeholder` in the event both are present

sophypal:
I ran into that before. You can't set a property in the options if it's already set. Or maybe it's the other way around.

sophypal:
No idea

